PR #30, opened 2019-05-18
Converting static paths to helper paths

### LINKED ISSUE #6: Make sure links all point to the correct guides and anchors

jwoertink:
fixes #6 

There's a lot of pages to do, but just want to make sure this is the path you were thinking. If this looks good, I'll finish the rest.

### DIFF
--- a/src/actions/guides/json_and_apis/saving-to-the-database.cr
+++ b/src/actions/guides/json_and_apis/saving-to-the-database.cr
@@ -1,5 +1,5 @@
 class Guides::JsonAndApis::SavingToTheDatabase < GuideAction
-  ANCHOR_SAVING_TO_THE_DATABASE = "saving-to-the-database_"
+  ANCHOR_SAVING_TO_THE_DATABASE = "perma-saving-to-the-database"
   guide_route "/json-and-apis/saving-to-the-database"
 
   def self.title
@@ -8,8 +8,7 @@ class Guides::JsonAndApis::SavingToTheDatabase < GuideAction
 
   def markdown
     <<-MD
-    <div id="#{ANCHOR_SAVING_TO_THE_DATABASE}"></div>
-
+    #{permalink(ANCHOR_SAVING_TO_THE_DATABASE)}
     ## Saving to the database
 
     Forms automatically know how to handle JSON params. They just need to be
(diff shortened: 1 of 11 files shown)


jwoertink:
Ok, that should be all of them.

jwoertink:
ok, I see what you're saying. That may make the markdown a bit messy throwing in random `<div>` tags, and trying to ensure those are closed in the right spot. Also, that means we need to add styling so those match up the same as the `# Heading`, or `## Sub Heading` ones.

paulcsmith:
I think the div tags would be empty so would not take any space. They’d just be there to link to and that’s it

paulcsmith:
And just to clarify the div wouldn’t replace the header. It would go just before it and would be empty. It’s purely there to link to

jwoertink:
Ok, I pushed up a single one just to test it. Let me know how that looks to you...

```crystal
Guides::JsonAndApis::SavingToTheDatabase.path(anchor: Guides::JsonAndApis::SavingToTheDatabase::ANCHOR_SAVING_TO_THE_DATABASE
```

A few small things. If you put the HTML right above the subheading, then the subheading is rendered as a literal 

```markdown
<div id="some-id"></div>
## subheading
```
renders
```html
<div id="some-id"></div>
## subheading
```

instead of

```html
<div id="some-id"></div>
<h2 id="subheading">subheading</h2>
```

So there has to be a space between them. And since the markdown will create an element with that ID, I have to add that extra `_` at the end to ensure they're different. 

If you like all that, then I'll do the rest of them.

paulcsmith:
Yeah I think making sure the links work is worth it. I've been to way too many guides with broken links and it sucks :(

One idea though to make sure these are consistent and have newlines so it doesn't break a heading:

We could add a method `permalink` to the `GuideAction` class that generates the div with a newline after:

```crystal
def permalink(anchor : String) : String
  %(<div id="#{anchor}"></div>\n)
end

"#{permalink(ANCHOR_NAME)}"
```

jwoertink:
I like that. I'll add that in and update the rest of them to match.

src/actions/guides/json_and_apis/saving-to-the-database.cr
```diff
@@ -1,4 +1,5 @@
 class Guides::JsonAndApis::SavingToTheDatabase < GuideAction
+  ANCHOR_SAVING_TO_THE_DATABASE = "saving-to-the-database_"
```
paulcsmith: And maybe rather than end with `_` it could be prefixed with `perma-`

jwoertink: That's fine too. That's just what will show up in the URL. Not sure if that's a factor when it comes to SEO or not.

paulcsmith: I think that is ok to show in the URL and hopefully won't affect SEO :P

jwoertink:
The conflict can take the version in master. Once this is in, I can go back and re-do a few of these. Plus I have some updated guides to push up so some in this PR will be irrelevant anyway.

PR #32, opened 2019-05-21
Updating http and routing docs

jwoertink:
* Updates some text for clarity
* Adds references to a ton more available methods
* Additional links and examples

### DIFF
--- a/src/actions/guides/http_and_routing/request_and_response.cr
+++ b/src/actions/guides/http_and_routing/request_and_response.cr
@@ -7,6 +7,46 @@ class Guides::HttpAndRouting::RequestAndResponse < GuideAction
 
   def markdown
     <<-MD
+    ## Handling Requests
+
+    When a request comes in to an Action, Lucky gives you access to the request object through the `request` method.
+    > The `request` object is an instance of [HTTP::Request](https://crystal-lang.org/api/HTTP/Request.html).
+
+    Lucky gives you access to a few helpful methods based on the request Content-Type.
+
+    * `json?` - true if the Content-Type header is "application/json"
+    * `ajax?` - true if the X-Requested-With header is "XMLHttpRequest"
+    * `html?` - true if the Content-Type header is "text/html"
+    * `xml?` - true if the Content-Type header is "application/xml" or "application/xhtml+xml"
+    * `plain?` - true if the Content-Type header is "text/plain"
+
+    You can use these methods to help direct the request or return different responses.
+
+    ```crystal
+    class Users::Show < BrowserAction
+      route do
+        if json?
+          # The Content-Type is a json request, so let's return some json
+          json(Users::ShowSerializer.new(current_user))
+        else
+          # Just render the page like normal
+          render Users::ShowPage
+        end
+      end
+    end
+    ```
+
+    ## Handling Responses
+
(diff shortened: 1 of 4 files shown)


src/actions/guides/http_and_routing/request_and_response.cr
```diff
@@ -7,6 +7,46 @@ class Guides::HttpAndRouting::RequestAndResponse < GuideAction
 
   def markdown
     <<-MD
+    ## Handling Requests
+
+    When a request comes in to an Action, Lucky gives you access to the request object through the `request` method.
+    > The `request` object is an instance of [HTTP::Request](https://crystal-lang.org/api/HTTP/Request.html).
+
+    Lucky gives you access to a few helpful methods based on the request Content-Type.
```
paulcsmith: Might be worth adding an example of using this in an action (like accessing headers)

Also might be worth mentioning you have access to the response object too but usually isn’t be used since we have helper methods. But can for fancier stuff

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -143,6 +143,8 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     end
     ```
 
+    > If the route requires an identifier for a specific resource, Lucky will generate a method by the name of that resource `_id`. (e.g. Users::Show generates user_id, and Projects::Show generates project_id)
```
paulcsmith: This part confused me a bit. Maybe something like this?

“Routed that generate an “id” param will prefix the peak with the underscores version of the resource name. “

And then the examples. Also I think the action and params should be wrapped in ` so it is highlighted as code

src/actions/guides/http_and_routing/request_and_response.cr
jwoertink: There's an example of these right below where they are [defined](https://github.com/paulcsmith/website-v2/pull/32/files#diff-6c5ae151a82b9614255254411efcd54aR25). 

I'll add a mention for the response object too. That could be handy.

jwoertink: oh.. I wonder if you meant to comment on the next section below. I just saw where I missed an example.

src/actions/guides/http_and_routing/routing_and_params.cr
jwoertink: hmm... Yeah, this one is tough. "peak" loses me here. Maybe
> Routes that require an "id" param will be prefixed with the resource name like `user_id`. 
And then show the examples?

paulcsmith: Stupid autocorrect. I meant prefix the param. But yeah I like that!

jwoertink: Ok, I pushed that up. Let me know if that reads ok. We can keep tweaking if it needs

src/actions/guides/http_and_routing/request_and_response.cr
paulcsmith: Oops. I meant to make this comment on the paragraph above that says you have access to the request. So you could show an example of accessing the request headers or something

jwoertink:
Ok, I added some examples of handling headers for both request and response. That should make a huge difference.

PR #38, opened 2019-05-28
Guides updates 3

jwoertink:
I broke these out in to separate commits, but these are all updates on what's available in lucky master. 

* fallback routing
* special tags like `nbsp`
* Secure headers
* structure change

### DIFF
--- a/src/actions/guides/frontend/rendering_html.cr
+++ b/src/actions/guides/frontend/rendering_html.cr
@@ -227,6 +227,18 @@ class Guides::Frontend::RenderingHtml < GuideAction
     There are some helpers for rendering HTML forms. For more info see the [saving
     data with forms](#{Guides::Database::ValidatingSavingDeleting.path(anchor: Guides::Database::ValidatingSavingDeleting::ANCHOR_USING_WITH_HTML_FORMS)}) guide.
 
+    ### Other special helpers
+
+    * `html_doctype` - Renders `<!DOCTYPE html>`
+    * `css_link(href, **options)` - Renders a `<link rel="stylesheet" media="sceen">` tag with `href` and any additional/override `options`
+    * `js_link(src, **options)` - Renders a `<script>` tag with `src` and any additional/override `options`
+    * `utf8_charset` - Renders a `<meta charset="utf8">` tag
+    * `responsive_meta_tag` - Another meta tag for responsive design.
+    * `nbsp(how_many)` - Renders `&nbsp;` entity for the number of times in `how_many` (1 by default).
+    * `raw` - Render RAW string to the page.
+
+    > Note: Using `raw` can be dangerous and should **never** be used with unescaped user-generated data.
+
     ## Rendering text
 
     Sometimes you want to render plain text. To do that use the `text` method.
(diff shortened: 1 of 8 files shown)


src/actions/guides/http_and_routing/http_handlers.cr
```diff
@@ -34,7 +35,7 @@ class Guides::HttpAndRouting::HTTPHandlers < GuideAction
     end
     ```
 
-    The `request` object will start in the `Lucky::HttpMEthodOverrideHandler`, do some processing, then move on to the `Lucky::LogHandler`, and so on.
+    The `request` object will start in the `Lucky::HttpMethodOverrideHandler`, do some processing, then move on to the `Lucky::LogHandler`, and so on.
```
paulcsmith: Should this change to ForceSSLHandler?

src/actions/guides/frontend/rendering_html.cr
```diff
@@ -227,6 +227,18 @@ class Guides::Frontend::RenderingHtml < GuideAction
...
     data with forms](#{Guides::Database::ValidatingSavingDeleting.path(anchor: Guides::Database::ValidatingSavingDeleting::ANCHOR_USING_WITH_HTML_FORMS)}) guide.
 
+    ### Other special helpers
+
+    * `html_doctype` - Renders `<!DOCTYPE html>`
+    * `css_link(href, **options)` - Renders a `<link rel="stylesheet" media="sceen">` tag with `href` and any additional/override `options`
+    * `js_link(src, **options)` - Renders a `<script>` tag with `src` and any additional/override `options`
+    * `utf8_charset` - Renders a `<meta charset="utf8">` tag
+    * `responsive_meta_tag` - Another meta tag for responsive design.
+    * `nbsp(how_many)` - Renders `&nbsp;` entity for the number of times in `how_many` (1 by default).
```
paulcsmith: Maybe `nbsp(how_many = 1)` so it is clear that there is a default even if you don't read the whole thing

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -211,7 +211,31 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+        if html?
+          render Home::IndexPage
+        else
+          raise Lucky::RouteNotFoundError.new(context)
+        end
+      end
+    end
+    ```
+
+    > ProTip: The `fallback` should always contain a `Lucky::RouteNotFoundError` error. This is to throw a 404 when an asset, or some other file is not found.
```
paulcsmith: We haven't used `ProTip` elsewhere so I'm nervous about introducing that. I think we can remove it and the visuals should show it is a tip because it looks different

src/actions/guides/http_and_routing/security_headers.cr
```diff
@@ -0,0 +1,77 @@
...
+    It's job is responsible for telling the browser to not render a page if it detects cross-site scripting.
+    Lucky disables this header for Internet Explorer version < 9 for you as per recommendations. Read more on [Microsoft](https://blogs.msdn.microsoft.com/ieinternals/2011/01/31/controlling-the-xss-filter/).
+
+    ```crystal
+    abstract class BrowserAction < Lucky::Action
+      include Lucky::SecureHeaders::SetXSSGuard
+    end
+    ```
+
+    ## Using HSTS
```
paulcsmith: Might be worth mentioning force ssl here too "Forcing SSL and HSTS" or something

src/actions/guides/http_and_routing/http_handlers.cr
jwoertink: haha. yup. Good catch.

PR #40, opened 2019-06-01
document page helpers

jwoertink:
This adds in docs for some of the pages helpers.

### DIFF
--- a/src/actions/guides/frontend/rendering_html.cr
+++ b/src/actions/guides/frontend/rendering_html.cr
@@ -267,23 +267,40 @@ class Guides::Frontend::RenderingHtml < GuideAction
 
     Formatting text on pages is pretty common. Lucky gives you several handy methods to help formatting.
 
+    > Some page helpers return a `String`, and some page helpers will write directly to the page.
+
     ### Converting a number to currency format
+
+    Returns a `String` formatted to a currency format.
 
     ```crystal
     # Returns standard U.S. format
-    number_to_currency(1234.43)
+    text number_to_currency(1234.43)
     # => $1234.43
 
     # Additional options supported for other formats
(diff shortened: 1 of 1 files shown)


src/actions/guides/frontend/rendering_html.cr
```diff
@@ -263,6 +263,95 @@ class Guides::Frontend::RenderingHtml < GuideAction
...
+
+    Use this to alternate between different values
+
+    ```crystal
+    array = [1, 2, 3]
+    cycle(array) #=> 1
+    cycle(array) #=> 2
+    cycle(array) #=> 3
+    cycle(array) #=> 1
+    ```
```
paulcsmith: Let's remove this for now. I'm not sure it is even useful and it seems fairly complex to implements so we might just remove it from Lucky

jwoertink: Agreed. That was sort of difficult to write lol

jwoertink:
Turns out this makes the docs a little bit more difficult. For example [this method](https://github.com/luckyframework/lucky/blob/master/src/lucky/page_helpers/text_helpers.cr#L101) contains HTML, but seems to return a String instead of writing to the page. 

I'm wondering if maybe those should just be broken out. Like one module that contains page helpers that return a flat string, and then a separate module of page helpers that write to the page....

In any case, I tried to make it clear in the docs which ones do what, so the wording may be a bit wonky. 

/cc. @edwardloveall

paulcsmith:
Yeah totally. I think there should be HtmlHelpers and TextHelpers so there is a clear file ration between those. Could you open an issue? I think that’d be a good cleanup task to make things nicer

edwardloveall:
Agreed, good distinction. Thanks for doing this @jwoertink !

PR #42, opened 2019-06-06
initial overhaul on database stuff

jwoertink:
Ok, I need to explain this one a bit. This is a "Proof of concept" on refactoring the database docs. I don't actually expect this PR to be merged, but more an initial discussion on what info we need to document, and where things should go. 

Here's a screenshot of the proposed sidebar:
Screen Shot 2019-06-06 at 2 07 39 PM

This proposal does a few things:

* Abstract the setup process for expansion in to fully configuring your database
* Abstract migrations to allow for more room to focus on this since it's a huge topic
* Lays out the structure for what avram uses (models, queries, forms, migrations)
* Adds an intro page explaining thoughts behind focus directly on postgres, some theory behind the avram name, and why it's different from other ORMs, plus a little section to show Lucky doesn't need Avram to run, so integrating other ORMs (i.e. for mysql, mongo, or using clear, etc..) would all be possible
* More in depth query guide show examples of things people would search for e.g. "lucky query A not eq B", etc...
* More in depth guide to using forms and how to properly use virtual forms
* Possibly removing HTML form stuff from this section and moving that to the frontend guide. The idea being by the time they get here, they've already covered HTML forms, so we can just gloss over most of that and focus on just the database related stuff.
* The raw SQL page should include using the non-type safe methods like `where`, and `order_by`
* The testing page would just add how to write a spec and test database related stuff.

Note that I deleted most of the actual docs and replaced with just placeholder text and notes. If we flesh out this structure, it'll make adding the actual docs a ton easier because we'll already know where things should go. 

* [x] - Intro page
* [x] - Setup page
* [x] - Migrations page
* [x] - Models page
* [x] - Queries page
* [ ] - Forms page
* [ ] - Raw SQL page
* [ ] - Boxes/Testing page

/cc. @edwardloveall


src/actions/guides/database/orms_and_avram.cr
```diff
@@ -0,0 +1,29 @@
+class Guides::Database::ORMsAndAvram < GuideAction
+  guide_route "/database/orms-and-avram"
+
+  def self.title
+    "ORMs and Avram"
```
paulcsmith: Maybe "Intro to Avram and ORMs"?

src/actions/guides/database/models.cr
```diff
@@ -7,104 +7,37 @@ class Guides::Database::Models < GuideAction
...
-        has_many tags : Tag, through: :taggings
-      end
-    end
-    ```
-
-    > The associations *must* be declared on both ends (the Post and the Tag in
-    this example), otherwise you will get a compile time error
+    ### Polymorphic?
 
+    ### Using STI?
```
paulcsmith: This doesn't really exist yet

src/actions/guides/database/models.cr
```diff
@@ -7,104 +7,37 @@ class Guides::Database::Models < GuideAction
...
+    ### Defining a column
 
-    > You can add a `?` to the end of any type to mark it as nilable.
+    show all options, and example of each supported type
 
-    ### Column types
+    Default generated columns
 
-    You can use the following types in your columns.
+    Nillable vs non-nillable types
```
paulcsmith: I love this 👍

src/actions/guides/database/models.cr
```diff
@@ -7,104 +7,37 @@ class Guides::Database::Models < GuideAction
 
   def markdown
     <<-MD
-    ## Setting up a model
+    ## Generate a Model
 
-    Let's create a `User` model with `lucky gen.model User` and then add some
-    columns. The name is a `String` and the age is an `Int32?`. The `age` column
-    allows null because the type is nilable (designated by the `?`).
+    Explain gen task with options. What files are created. Each section should link to page explaining those (i.e. ModelQuery links to queries, Form links to saving, migration links to migrations page).
```
paulcsmith: Love the idea.

Also should have a brief intro on what a model is.

src/actions/guides/database/models.cr
```diff
@@ -7,104 +7,37 @@ class Guides::Database::Models < GuideAction
...
-    Right now you can define `has_many`, `has_one`, and `belongs_to` associations.
+    > Note that adding require statements in to the model may be necessary due to how crystal require works?
 
-    ```crystal
-    # Require the models we want to associate. If we don't we may get an
-    # "Undefined constant" error because Lucky didn't load the associated models yet
-    require "./supervisor"
-    require "./task"
-    require "./company"
+    ### Belongs to
```
paulcsmith: Also link to relevant sections in migrations too

src/actions/guides/database/models.cr
```diff
@@ -7,104 +7,37 @@ class Guides::Database::Models < GuideAction
...
-        column title : String
-        has_many taggings : Tagging
-        has_many tags : Tag, through: :taggings
-      end
-    end
-    ```
-
-    > The associations *must* be declared on both ends (the Post and the Tag in
-    this example), otherwise you will get a compile time error
+    ### Polymorphic?
```
paulcsmith: Not yet officially supported in an easy to use way but can be done in a more custom way. We can maybe add something later as this might take a bit more trial and error to figure out a good suggestion

src/actions/guides/database/orms_and_avram.cr
```diff
@@ -0,0 +1,29 @@
...
+
+    A note about the name "Avram". Link to the Avram repo. Lucky decision to focus on Postgres.
+
+    ## Parts to Avram
+
+    * Models
+    * Queries
+    * Forms
+    * Boxes
+    * Migrations
```
paulcsmith: I love this 👍 Maybe they are each their own headline though so they get shown in the table of contents

src/actions/guides/database/querying.cr
```diff
@@ -7,293 +7,79 @@ class Guides::Database::Querying < GuideAction
 
   def markdown
     <<-MD
-    ## Querying the database
+    ## Query Objects
+    
+    How Lucky uses these instead of query on the model directly. The structure of these and where they go in the app. Type safety
```
paulcsmith: Helpful to explain that model columns generate methods on the query object

src/actions/guides/database/querying.cr
```diff
@@ -7,293 +7,79 @@ class Guides::Database::Querying < GuideAction
...
-    ```
+    ### A != B
 
-    ### Customizing how associations are preloaded
+    ### A gt/lt B
 
-    Sometimes you want to order preloads, or add where clauses. To do this, use
-    the `preload_{{ association_name }}` method on the query and pass a query
-    object for the association.
+    ### A in / not in (B)
```
paulcsmith: I love these specific headings. Nice

src/actions/guides/database/querying.cr
```diff
@@ -7,293 +7,79 @@ class Guides::Database::Querying < GuideAction
...
 
-    ```crystal
-    post = PostQuery.first
-    # Returns the associated author and does not trigger a preload error
-    post.author!
-    ```
+    ## Associations and Joins
 
-    ## Joining and querying associations
+    ### Preloading
```
paulcsmith: Preloading should probably be `##` because it is a pretty big/important concept that could use a lot more examples and explanations

src/actions/guides/database/querying.cr
```diff
@@ -7,293 +7,79 @@ class Guides::Database::Querying < GuideAction
...
-        # See "Short one-argument syntax" in https://crystal-lang.org/docs/syntax_and_semantics/blocks_and_procs.html
-        comments(&.rating.gte(4))
-      end
-    end
+    Explain `none`
 
-    # Or do the same without making a scope method:
-    PostQuery.new.join_comments.comments(&.rating.gte(4))
-    ```
+    ## Scopes
```
paulcsmith: Maybe `Named Scopes`

src/actions/guides/database/testing.cr
```diff
@@ -9,6 +9,8 @@ class Guides::Database::Testing < GuideAction
     <<-MD
     ## Creating a Box
 
+    More docs on how this all works. Show examples used in specs.
```
paulcsmith: 👍

src/actions/guides/database/validating-saving-deleting.cr
```diff
@@ -8,676 +8,64 @@ class Guides::Database::ValidatingSavingDeleting < GuideAction
...
-    # src/forms/user_form.cr
-    class UserForm < User::BaseForm
-    end
-    ```
+    ## Creating Records
 
-    > You can have more than one form object to save a record in different ways.
-    For example, default Lucky apps have a `SignUpForm` specifically for handling
-    user sign ups.
+    Examples of passing in `params`, or how it would look being used in a task (i.e. seeing).
```
paulcsmith: Yeah and I like the idea of showing how to do it without params first, and then with them later. That way it is clear you pass in named args

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,39 @@
...
+
+  def self.title
+    "Database Setup"
+  end
+
+  def markdown
+    <<-MD
+    ## Configure
+
+    Maybe a mention of which version of postgres needs to be installed or min version?
```
paulcsmith: I think most recent versions work but I am not sure of an exact version. Prob v9, but depends on what you need more than Avram. Like if you need JSONB I think that is v9

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,39 @@
...
+
+    Mention `config/database.cr` and how to configure connection and Avram configure options
+
+    > Note about CLI tools for postgres app on macOS.
+
+    ### Test setup
+
+    Notes about using .env so test DB doesn't conflict with dev DB
```
paulcsmith: This is done with `config/database.cr` and not `.env` Might be worth mentioning that.

Also might want to add a section called `Setting local credentials` in which case you'd want to use `config/local/database.cr` where you can override the settings. Or use `.env`

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,39 @@
...
+
+    ## Create or Drop
+
+    How to run create task
+
+    How to run drop task
+
+    ## Seeding Database
+
+    Explain differences in seeding tasks
```
paulcsmith: Great idea

paulcsmith (commented):
I freakin love it. Thanks for tackling this. I had some comments on a few spots, but mostly small stuff. If you wanna start filling these out that'd be awesome. Feel free to ping on sections you want comments on or would like help with

src/actions/guides/database/validating-saving-deleting.cr
jwoertink: Yeah, I'd like to show an example of what the params look like. So you know when you pass in `params`, it looks like `{"user:name" => "whatever"}` or however that actually looks. I had to boot an app, make a form, and then inspect just to figure out what that looks like 😅

jwoertink:
What are your thoughts on changing the main title of the section from `Database` to something like `Database and Models`?

src/actions/guides/database/validating-saving-deleting.cr
paulcsmith: The params are a bit tricky though because you probably should never pass them in yourself in that way. It's almost exclusively for HTML form params/JSON params. If you need to do something not user generated it should pretty much always use the named args for type safety. Maybe that is good to include too?

Like an example showing parsing a simple CSV and then using that in a form:

```crystal
# Total pseudo code. No idea if this is even in the std lib
csv_row = Csv.parse("Paul, Smith").first # Get the first row only
UserForm.create!(first_name: csv_row.first, last_name: csv_row.last)
```

edwardloveall:
Overall I'm a fan of this whole idea, both how your presenting it here in the PR and also the changes you've made. I'm pretty slammed with other responsibilities right now to take a close look. If you can wait a couple of weeks I'd love to revisit. If not, carry on! I trust you both.

paulcsmith:
I think this can be closed, but feel free to re-open if I was wrong!

jwoertink:
It's ok. I think after this next release, all these pages will need some more updates anyway 😅

PR #49, opened 2019-06-08
adding in the second page for the new database section

jwoertink:
This adds the new DatabaseSetup page.


src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+    <<-MD
+    ## Configure
+
+    Ensure that your system is running a [PostgreSQL](https://www.postgresql.org/docs/current/tutorial-install.html) version of at least 9.4 or later.
+
+    > If you're using the postgres app on macOS, be sure to install [CLI tools](https://postgresapp.com/documentation/cli-tools.html).
+
+    ### Standard Options
+
+    To configure Lucky to connect to your database, you can define a `DATABASE_URL` environment variable. (e.g. `postgres://root@localhost:5432/my_database`). To see more options, look in your app's `config/database.cr` file.
```
paulcsmith: I think DATABASE_URL should come after the config/database.cr explanation. Since the config file is much easier to use. The DATABASE_URL could be explained right after the config options in this section

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+    Lucky also supports connection pool settings:
+
+    * initial_pool_size
+    * max_pool_size
+    * max_idle_pool_size
+    * checkout_timeout
+    * retry_attempts
+    * retry_delay
+
+    To set the connection pool options, just append a query string to the end of the `settings.url` option (e.g. `?initial_pool_size=5&retry_attempts=2`)
```
paulcsmith: I love this!

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+
+    ### Avram Configuration
+
+    Avram requires a `url` option to be set. If you decide to not use Avram as your ORM, you can set this option to a string like `"noop"`.
```
paulcsmith: I’ve seen lots of people get confused by “noop” (reading it like poop 💩). Thoughts on “unused”?

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+
+    Optionally, the `lazy_load_enabled` is set to `false` for development and test. This causes Lucky to raise an exception if you forget to preload an association.
```
paulcsmith: Might be helpful to add “but will not raise an error in production”

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+
+    ### Drop database
+
+    To drop the database, run the `lucky db.drop`
+
+    > Please *please* don't ever do this in production
+
+    ## Seeding Data
+
+    Seeding is the process of putting initial data in to your database. This could be fake placeholder data you use in development, or even special data your application expects to exist in production.
```
paulcsmith: I think we should remove “initial”. Because it can really be used afterward or anytime. I’ve seen people run it on every deploy too. So I think “putting data into your database” should work for all cases

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+
+    By default, Lucky generates two tasks in your app's `tasks/` folder. `Db::CreateRequiredSeeds`, and `Db::CreateSampleSeeds`. You can use [Boxes]() or [Forms]() to create the data. Then you will place this code in the `call` method of those tasks
```
paulcsmith: I’m not sure I understand what this means by “place this code”. Maybe the code was not committed? Or maybe we can remove that line?

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+    ### Required Seeds
+
+    Let's say you're getting ready to launch your application to production for the very first time. You may need an initial Admin user account that will be able to login and create your other Admin accounts.
+
+    This code will go in `tasks/create_required_seeds.cr`.
+
+    ```crystal
+    def call
+      # Using a Box
+      UserBox.create &.email("developer@example.com").status(User::Status::Admin.value)
```
paulcsmith: I like using enum but it might be too much Crystal knowledge for beginners. What do you think of using `admin: true` to keep it super simple?

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+
+      # Using a Form
+      UserForm.create!(email: "developer@example.com", status: User::Status::Admin.value)
+    end
+    ```
+
+    Run this task with `lucky db.create_required_seeds`.
+
+    > This task should be ran after your first deployment of your application.
```
paulcsmith: Maybe “you should run this after your first deployment, and whenever the seeds change”

src/actions/guides/database/database_setup.cr
```diff
@@ -0,0 +1,113 @@
...
+      # Using a Form
+      100.times do |i|
+        ProductForm.create(name: "Product \#{i}")
+      end
+    end
+    ```
+
+    Run this task with `lucky db.create_sample_seeds`.
+
+    > Running `./script/setup` in development will run the `db.create_sample_seeds` task for you. If you need to re-seed, you can run `lucky db.drop` and then `./script/setup` to re-create and seed your local database.
```
paulcsmith: We also run the required seeds task so we should probably mention that in the previous section

paulcsmith:
Looks like it needs a rebase. But other than that looks great

PR #52, opened 2019-06-09
adding in the third page (migration docs) for the new database section

jwoertink:
All the new migration docs.


src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+    <<-MD
+    ## Migrating Data
+
+    ### Introduction
+
+    You can think of migrations like version control for your database. These are classes that are scoped with a timestamp to allow you to update your database in versions as well as undo changes that you've made previously while keeping track of the order you made them in.
+
+    ### Generating a new one
+
+    To create a new migration, run the `lucky gen.migration {PurposeOfMigration}`. (e.g. If you need to add a `name` column to `users`, you would run `lucky gen.migration AddNameToUsers`). This would generate a `db/migrations/#{Time.utc.to_s("%Y%m%d%H%I%S")}_add_name_to_users.cr` file.
```
paulcsmith: I really like this explanation!

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+
+    > When you generate a new model, a migration file is created for you.
+
+    ### Anatomy
+
+    A new migration file will always start with a timestamp. This is to keep the order in which they are ran. You may have a migration that adds a column, and then a year later, another one that removes that column. You want to run these in the correct order.
```
paulcsmith: Maybe file -> file name to make it clearer. Love the explanation of why this is important!

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+
+    In that file, you'll find two methods `migrate`, and `rollback`.
+
+    * migrate - Use this method to create your migration code.
```
paulcsmith: This description didn’t really click for me. Maybe “this code will run when moving forward (migrating)”. Im not sold on this but maybe gives you some ideas?

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+    * rollback - This method should un-do everything the `migrate` method does.
```
paulcsmith: Perfect explanation 👌 

I think un-do can be one word without the hyphen though

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+
+    > The `rollback` method should do the opposite of `migrate` in reverse order. (e.g. If `migrate` adds a column, then adds an index for that column, `rollback` should remove the index first, then remove the column.)
```
paulcsmith: Good example!

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+
+    ```crystal
+    def migrate
+      create :users, primary_key_type: :uuid do
+        add email : String
+      end
+    end
+    ```
+
+    > Be sure to update your model with `primary_key_type: :uuid` as well.
```
paulcsmith: Maybe prefix with: “if you do this, “

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+    def migrate
+      drop :users
+    end
+    ```
+
+    > If your `migrate` method runs `create`, then the `rollback` method should run `drop`.
+
+    ## Alter table
+
+    If your table exists, but you need to make changes to it, use the `alter` method. All of your changes will go in the `alter` block.
```
paulcsmith: Maybe “ your changes *to this table*”?

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+    def migrate
+      alter :users do
+        remove :old_field
+      end
+    end
+    ```
+
+    ## Add column
+
+    To add a new column, you'll use the `add` method inside of a `[create](#create-table)` block.
```
paulcsmith: Or alter table

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+        # set to the users table.
+        #
+        # When the associated author is deleted, their comments are also deleted
+        # because we set on_delete: :cascade
+        add_belongs_to author : User, on_delete: :cascade
+      end
+    end
+    ```
+
+    You must include the `on_delete` option which can be one of `:cascade`, `:restrict`, `:nullify`, or `:do_nothing`.
```
paulcsmith: I think we should explain these options. Some context here https://github.com/luckyframework/lucky/issues/773#issuecomment-491416449

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+    To add a new column, you'll use the `add` method inside of a `[create](#create-table)` block.
+
+    ```crystal
+    create :users do
+      add email : String
+      add birthdate : Time
+      add login_count : Int32
+    end
+    ```
+
```
paulcsmith: I think a section on fill_existing_with would be helpful. Basically if you have an alter table with a required column it requires you to use fill_existing with. That can be a value, or :do_nothing

The idea is that when you add a required column to an existing table with rows it will fail. So we make it easy to fill existing columns with a default before applying the non null constraint. 

You can use :do_nothing though if the rows are empty or if you plan to handle data migration later

src/actions/guides/database/migrations.cr
```diff
@@ -0,0 +1,231 @@
...
+    ```crystal
+    def migrate
+      drop_index :users, [:status]
+    end
+    ```
+
+    ## Associations
+
+    If your tables have a one-to-many or a many-to-many relation, you can use these methods to create a foreign key constraint, or remove that constraint from your tables.
```
paulcsmith: This may be if some help too: https://github.com/luckyframework/lucky/issues/773#issue-442184413

Maybe we can explain this with some examples? I’m fine shipping this as-is but just wanted to reference that in case you’ve got some ideas

PR #64, opened 2019-06-13
adding the forth page (models) for the new database section

jwoertink:
This one can for sure be better even still, but this adds in a lot of new things that aren't on the current guides.

One thing to note is I want to link to the new pages, but they all need to be merged in. My thought was get all these pages merged in, then do a last once over on them to get all the proper links between the pages in place.


src/actions/guides/database/models.cr
```diff
@@ -7,70 +8,160 @@ class Guides::Database::Models < GuideAction
 
   def markdown
     <<-MD
-    ## Setting up a model
+    ## Introduction
+
+    A Model is an object used to map a corresponding database table to a class. These objects model real-world objects to give you a better understanding on how they should interact within your application.
+
+    Models in Lucky allow you to define methods associated with each column in the table. These methods return the value set in that column.
```
paulcsmith: I think another paragraph saying `Models also generate classes you can use to save new records, and query existing ones` or something like that

src/actions/guides/database/models.cr
```diff
@@ -7,70 +8,160 @@ class Guides::Database::Models < GuideAction
...
-    > Note: we'll assume we ran a migration that created a name and age column.
+    Once you run the model generator, you'll have a file that looks like this
 
     ```crystal
     # src/models/user.cr
     class User < BaseModel
       table :users do
-        column name : String
-        column age : Int32?
+        # You will define columns here
```
paulcsmith: I kind of like having the columns, but agree it is odd to just have them there randomly. Maybe `# You will define columns here. For example:` and then a couple columns. What do you think?

src/actions/guides/database/models.cr
```diff
@@ -7,70 +8,160 @@ class Guides::Database::Models < GuideAction
...
+
+    ### Adding a column
+
+    Inside of the `table` block, you'll add the columns your model will define using the `column` method.
+
+    ```crystal
+    table :users do
+      column email : String
+      column active : Bool
+      column ip_address : String?
```
paulcsmith: I think it'd be valuable to add a comment here saying `This type can be nil because it ends with '?'`. Even though it's mentioned earlier I think having it here would help

src/actions/guides/database/models.cr
```diff
@@ -7,70 +8,160 @@ class Guides::Database::Models < GuideAction
...
+    ```crystal
+    table :users do
+      column name : String
 
+      # gives you the `company_id`, and `company` methods
+      belongs_to company : Company
+    end
+    ```
+
+    > When you create the migration, be sure you've set `[add_belongs_to](#{Guides::Database::ManagingAndMigrating.path})`.
```
paulcsmith: Nice. This is merged now so you should be able to anchor to it directly

src/actions/guides/database/models.cr
```diff
@@ -7,70 +8,160 @@ class Guides::Database::Models < GuideAction
...
     UserQuery.new.preload_company.find(1).company
     ```
 
-    ### Making associations optional
+    ### Optional association
 
     Sometimes associations are not required. To do that add a `?` to the end of the type.
 
-    ```crystal
+     ```crystal
```
paulcsmith: Extra space snuck in here

src/actions/guides/database/models.cr
```diff
@@ -7,70 +8,160 @@ class Guides::Database::Models < GuideAction
...
-    > Make sure to make the column nilable in your migration as well: `belongs_to Company?`
+    > Make sure to make the column nilable in your migration as well: `add_belongs_to company : Company?`
 
-    ### Has many through (many-to-many)
+    ## Has one (one to one)
+
+    ```crystal
+    table :users do
+      has_one supervisor : Supervisor
+    end
```
paulcsmith: I believe this should also show the other direction (`belongs_to`) so it is clear how the relationship works.

src/actions/guides/database/models.cr
```diff
@@ -84,15 +175,13 @@ class Guides::Database::Models < GuideAction
         belongs_to post : Post
       end
     end
-
```
paulcsmith: I thin having a newline between the classes adds a nice separation. What do you think about adding them back?

PR #66, opened 2019-06-13
Adding in full CORS example. 

### LINKED ISSUE #46: Document CORS handler
https://lucky-website.herokuapp.com/guides/json-and-apis/cors

@jwoertink I think you've got some experience with this

jwoertink:
Fixes #46

CORS is confusing... I don't think these docs really make it "clear", but they do provide a full example that could technically be dropped right in to an app.

### DIFF
--- a/src/actions/guides/json_and_apis/cors.cr
+++ b/src/actions/guides/json_and_apis/cors.cr
@@ -36,9 +36,8 @@ class Guides::JsonAndApis::Cors < GuideAction
     ### Complex example
 
     You may find that you need a little more than a simple setup to handle CORS.
-    Some requests may require a preflight `OPTIONS` call. We have two ways we can handle this.
-    We can create a `match :options, "/api/"` route block for every action in our API, or move this
-    to a handler.
+    Some requests may require a preflight `OPTIONS` call. To best handle this,
+    we will create a new [handler](#{Guides::HttpAndRouting::HTTPHandlers.path}).
 
     Add a new folder called `src/handlers/`, and be sure to add a require in `src/app.cr`.
     We can place our new `CORSHandler` in `src/handlers/cors_handler.cr`.
--- a/src/actions/guides/json_and_apis/cors.cr
+++ b/src/actions/guides/json_and_apis/cors.cr
@@ -11,6 +11,8 @@ class Guides::JsonAndApis::Cors < GuideAction
 
     When working with an API, you may need to set some [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) headers. Doing this in Lucky is pretty easy!
 
+    ### Basic example
+
     In your `src/actions/api_action.cr`, you can use a `before` action to set your headers.
 
     ```crystal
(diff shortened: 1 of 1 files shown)


src/actions/guides/json_and_apis/cors.cr
```diff
@@ -21,12 +23,109 @@ class Guides::JsonAndApis::Cors < GuideAction
...
       end
     end
     ```
 
     > Note: the before action method must return `continue`, or a `LuckyWeb::Response`. See [Actions and Routing](#{Guides::HttpAndRouting::RequestAndResponse.path(anchor: Guides::HttpAndRouting::RequestAndResponse::ANCHOR_RUN_CODE_BEFORE_OR_AFTER_ACTIONS_WITH_PIPES)}) for more info.
+
+    ### Complex example
+
+    You may find that you need a little more than a simple setup to handle CORS.
+    Some requests may require a preflight `OPTIONS` call. You can check for that request method.
```
paulcsmith: How does this work? Do you need to specify an `options` route for every route? It seems like if you didn't it would never hit this `before` action because there'd be no matching requests?

But also I don't know a ton about CORS so maybe I just don't get it :P

jwoertink: @russ ^ ??? 😂

russ: Bit of a miscommunication there. The example needs to be it's own handler inserted into middleware stack before the RouteHandler. That way CORS is handled for the whole app as opposed to each action.

jwoertink: Ok, that makes more sense. I'll move it to a handler

src/actions/guides/json_and_apis/cors.cr
```diff
@@ -21,12 +23,167 @@ class Guides::JsonAndApis::Cors < GuideAction
...
 
-    > Note: the before action method must return `continue`, or a `LuckyWeb::Response`. See [Actions and Routing](#{Guides::HttpAndRouting::RequestAndResponse.path(anchor: Guides::HttpAndRouting::RequestAndResponse::ANCHOR_RUN_CODE_BEFORE_OR_AFTER_ACTIONS_WITH_PIPES)}) for more info.
+    This will set these headers before every action in your API.
+
+    > Note: the `before` pipe must return `continue`, or a `LuckyWeb::Response`. See [Actions and Routing](#{Guides::HttpAndRouting::RequestAndResponse.path(anchor: Guides::HttpAndRouting::RequestAndResponse::ANCHOR_RUN_CODE_BEFORE_OR_AFTER_ACTIONS_WITH_PIPES)}) for more info.
+
+    ### Complex example
+
+    You may find that you need a little more than a simple setup to handle CORS.
+    Some requests may require a preflight `OPTIONS` call. We have two ways we can handle this.
```
paulcsmith: I think we can remove the part about `match :options` since it is so inflexible. I bet the handler is the right call almost every time

jwoertink:
I think it's a toss up. On the one hand, it's nice to show that we can handle it in the actions pretty easy, but on the other hand when it comes to CORS, you're most likely to go all or none situation. 

I'm good either way with it.

paulcsmith:
If you're cool with it, let's go with the handler approach. Just to make sure that it always works with everything. Maybe we can extract some helper methods in Lucky later that make the CORS handler easier to use, but for now I think this is great

PR #71, opened 2019-06-21
Adding in new querying database page

jwoertink:
This is sort of a gnarly page because there's so much to explain. I'm pushing up now so we can get the review going, but I'll start pushing up those changes as well as new stuff that's near the bottom of the page.


paulcsmith:
Note to anyone curious about this PR: Check out https://lucky-website-pr-71.herokuapp.com/guides/database/querying to see this live. Might be easier to understand if you look there first

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,240 @@ class Guides::Database::Querying < GuideAction
...
-    ## Ensuring no results are returned
+    Select distinct rows based on the `name` column `SELECT DISTINCT ON (users.name) FROM users`
 
     ```crystal
-    UserQuery.new.none
+    UserQuery.new.distinct_on(&.name)
     ```
 
-    This can be helpful when under certain conditions you want the resultd to be empty.
+    > Notice these two methods are called in the instance of `UserQuery`, and not the class.
```
paulcsmith: This isn't *quite* right actually all the methods are instance methods and maybe that's what should be documented here. The `first/last/etc` on the class are shortcuts that call `new.first/last/etc`. So I'd suggest we use those and add a separate section `###` that says `Select shortcut` methods.

May also be worth mentioning that you can chain these like `UserQuery.new.age(30).first`

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,240 @@ class Guides::Database::Querying < GuideAction
...
-    to actions and the model to make this easy and maintainable.
-
-    ## Ordering results
-
-    ````crystal
-    UserQuery.new.age.desc_order
-    # or
-    UserQuery.new.age.asc_order
-    ````
+    ### Select one
```
paulcsmith: I think splitting these into `Select first`, `select last` and `select by primary key (find)`. That way they will show up in search better (the heading will be what people search for. May also be worth noting there are `?` methods of these and that if you use the non-question mark version it will raise if it doesn't exist.  Using `?` will return `nil` if it doesn't exist

You do this above but I think a note here would be good as well in case they get here by a search and don't see the previous sections

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,240 @@ class Guides::Database::Querying < GuideAction
...
+    user = UserQuery.find(3)
     ```
 
-    ## Finding by ID
+    ### Select distinct / distinct on
 
-    ```crystal
-    id = 1
-    UserQuery.new.find(id)
+    `SELECT DISTINCT COLUMNS FROM users`
```
paulcsmith: I love that you show the SQL 😍

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,240 @@ class Guides::Database::Querying < GuideAction
...
+    UserQuery.new.name.not.eq("Billy")
+    ```
+
+    > The `not` method can be used to negate other methods like `eq`, `gt`, `lt`, `in`.
 
-    * `gt`
-    * `gte`
-    * `lt`
-    * `lte`
+    ### A gt/lt B
```
paulcsmith: It may be good to add a short sentence between the SQL and the heading that explains this in human readable terms so it is searchable. Something like `Where column A is greater than or less than column B`. That kind of thing

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
-    ## Where clauses
+    ### Select first
 
-    You can use the methods generated by Lucky to create where clauses.
+    `SELECT COLUMNS FROM users LIMIT 1`
 
     ```crystal
-    # SELECT * FROM users where name='Paul';
-    UserQuery.new.name("Paul")
+    # raise Avram::RecordNotFound if nil
```
paulcsmith: I love the inline examples 👍

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
 
-    ## Ordering results
+    ```crystal
+    UserQuery.new.distinct_on(&.name)
+    ```
 
-    ````crystal
-    UserQuery.new.age.desc_order
-    # or
-    UserQuery.new.age.asc_order
```
paulcsmith: This is the type sage way of ordering. This should be the main way to order i  the “ordering” section. The order_by method is the non type safe but more flexible way since you can do string interpolation and stuff. It should be avoided if possible though

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
+    ### Avg / Sum
 
-    Sometimes you want to order preloads, or add where clauses. To do this, use
-    the `preload_{{ association_name }}` method on the query and pass a query
-    object for the association.
+    `SELECT AVG(users.age) FROM users`
 
     ```crystal
-    PostQuery.new.preload_comments(CommentQuery.new.published(true))
+    UserQuery.new.select_average(:age)
```
paulcsmith: I think you can use a block like: select_avg(&.age) for type safety. But if not then we should add that to Avram hah

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
-    * `left_join{{association_name}}`
+    If you need to just delete every record in the entire table, you can use `destroy_all` to truncate.
 
-    So if `Project` `has_many tasks : Task` you'd have `join_tasks`,
-    `inner_join_tasks`, etc.
+    `TRUNCATE TABLE users`
 
+    ```crystal
+    UserQuery.new.destroy_all
+    ```
```
paulcsmith: Might be worth showing an example showing that you can also scope this. UserQuery.new.archived_at.lte(30.days.ago).delete_all

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
 
-    If we want to get all posts that have at least 1 highly rated comment, we
-    could do something like this:
+    ## Preloading
+
+    ### without preloading
+
+    ## No results
+
+    Avram gives you a `none` method to return no results. This can be helpful when under
```
paulcsmith: Resultd should be results

src/actions/guides/database/querying.cr
paulcsmith: Actually it is `column.select_min/max`. See https://github.com/luckyframework/avram/blob/5d7a5090d37a702ff7b7f5b59fb602e4d29dc6e5/spec/query_spec.cr#L312-L382

src/actions/guides/database/querying.cr
jwoertink: I don't think that would work. https://github.com/luckyframework/avram/blob/master/src/avram/query.cr#L6 This method just truncates the whole DB. Or is there another definition somewhere?

paulcsmith: Hmmm, that is not expected. I will open an issue about this. I would expect there to be a way to delete with a where

src/actions/guides/database/querying.cr
jwoertink: Ah, is that not these methods https://github.com/luckyframework/avram/blob/master/src/avram/query_builder.cr#L123-L137 then?

src/actions/guides/database/querying.cr
paulcsmith: That method should probably be renamed `truncate` to be super clear https://github.com/luckyframework/avram/issues/110

paulcsmith: And https://github.com/luckyframework/avram/issues/111

paulcsmith: But this looks good for now. Might be worth mentioning it isn't chain able and will truncate the table so people aren't surprised

src/actions/guides/database/querying.cr
paulcsmith: Oops I wasn't clear. What you documented is correct and will work, but there are type-safe versions that should be preferred. Those are the ones I linked to. The method definitions arehttps://github.com/luckyframework/avram/blob/8805b31be56f39976fa8264f3e99aec9fd51b239/src/avram/criteria.cr#L63-L81

Most of the non-type safe versions are available but should only be used if your query is doing something out of the ordinary. The type-safe ones call the non type-safe ones under the hood, but since they are called on a Column's criteria we know the column exists

jwoertink: ah. Ok. In that case, maybe I should remove the docs on the `where` that uses the non-type safe call.

paulcsmith: Actually I think that one is handy in case you need something Avram doesn't support. Like if you need to do a substring match or one of the JSON methods or something. But maybe it belongs in the `Raw SQL` section and the section here should just link to that `### Raw where` -> `"Sometimes Avram doesn't give you exactly what you need. Take a look at the Raw SQL guide to learn how to write any query you need"`

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
+    ### Custom WHERE
 
-    * `gt`
-    * `gte`
-    * `lt`
-    * `lte`
+    Avram also has a `where` method that gives you the ability to do custom and/or programmatic queries.
 
-    ### Examples
+    `WHERE users.name = 'Dan' AND created_at > '#{3.days.ago}' AND created_at < '#{1.hour.ago}'`
```
edwardloveall: I'd be great to get an example here that requires a custom where. For example:

```sql
WHERE name SIMILAR TO '([AEIOU])%'
```

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
-    If we want to get all posts that have at least 1 highly rated comment, we
-    could do something like this:
+    ## Preloading
+
+    ### without preloading
+
+    ## No results
+
+    Avram gives you a `none` method to return no results. This can be helpful when under
+    certain conditions you want the resultd to be empty.
```
edwardloveall:
```suggestion
    certain conditions you want the results to be empty.
```
What @paulcsmith said

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
-        comments { |comment_query| comment_query.rating.gte(4) }
+    > This method does not return an empty array immediately. You can still chain other query methods,
+    > but it will always return no records. For example: `UserQuery.new.none.first` will never return a result
 
-        # Or do the same thing with the block shorthand syntax
-        # See "Short one-argument syntax" in https://crystal-lang.org/docs/syntax_and_semantics/blocks_and_procs.html
-        comments(&.rating.gte(4))
+    ## Named Scopes
+
+    Generally it’s best to name your query methods so they are easy to understand and reuse across your app.
```
edwardloveall:
```suggestion
    Chaining multiple query methods can be hard to read, tedious, and error prone. If you are making a complex query more than once, or want to give a query a label, named scopes are a great alternative.
```

Sorry didn't want to just clobber your thing here, I just thought it could use a bit more explanation of _why_ named scoped should be used.

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
     ```crystal
-    UserQuery.new.distinct_on(&.name)
+    users = UserQuery.new.all
     ```
 
-    ## Where clauses
+    ### Select first
 
-    You can use the methods generated by Lucky to create where clauses.
+    `SELECT COLUMNS FROM users LIMIT 1`
```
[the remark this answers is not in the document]
jwoertink: I don't think so https://github.com/luckyframework/avram/blob/master/src/avram/queryable.cr#L96-L97

It just adds a `limit(1)`, and then calls the `.first?` method from Array. Maybe it should?

src/actions/guides/database/querying.cr
jwoertink: good idea

src/actions/guides/database/querying.cr
jwoertink: haha. no worries! Thanks for the help!

src/actions/guides/database/querying.cr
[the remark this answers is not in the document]
edwardloveall: Mmm I see what's happening. I was confused as to why `first` didn't have ordering but `last` did. I think it's because `last` adds `reverse_order` which sets an order by clause. So, not a problem here, but I'll put an issue in on Avram.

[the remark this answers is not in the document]
edwardloveall: https://github.com/luckyframework/avram/issues/115

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
+    `WHERE users.age >= 21`
+
+    ```crystal
+    UserQuery.new.age.gte(21)
     ```
 
-    ## Finding by ID
+    Find rows where `A` is greater than `B`.
+
+    `WHERE users.created_at > '#{1.day.ago}'`
```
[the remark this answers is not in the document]
jwoertink: I do this in a few guides, because I'm *that* guy that would be like "oh, this is out of date"... 😂

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,428 @@ class Guides::Database::Querying < GuideAction
...
+    ```
 
-    # The query will run once `each` is called
-    # Results are not cached so a request will be made every time you call `each`
-    query.each do |user|
-      pp user.name
-    end
+    > The `not` method can be used to negate other methods like `eq`, `gt`, `lt`, `in`.
+
+    ### A gt/lt B
```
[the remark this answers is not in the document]
jwoertink: Yeah, it is sort of a lot to take in. The benefit this way though is being able to search for them individually. Like searching "less than". I'm not sure how that search works though. Maybe it would pick up the code examples if I used crystal comments?

```crystal
# less than or equal to `WHERE users.age <= 12`
UserQuery.new.age.lte(12)

# less than `WHERE users.updated_at < '#{3.months.ago}'`
UserQuery.new.updated_at.lt(3.months.ago)
```

src/actions/guides/database/querying.cr
jwoertink: I think this is going to get moved to the raw SQL page. I'm adding a note about it.

jwoertink:
Latest push I've started adding in join docs. One thing the current docs has that kind of confused me is this
```crystal
def with_highly_rated_comments
    join_comments

    comments { |comment_query| comment_query.rating.gte(4) }
end
```

Is there just a `.` missing here? Or do you not have to chain these methods for them to work?


Also, I think I want to move the raw where and those less type-safe methods over to the custom SQL page. That way we can talk about how you have the ability to do custom things with chainable methods, but the type-safety isn't as good.

paulcsmith:
You have to chain them outside of instance methods. But inside an instance method it will operate on the instance of the class so you don’t need to chain. Maybe better to chain them anyway though just to keep things consistent! They can still be in separate lines though I think

join_comments
.comments etc.

And yeah I think moving the less type safe methods to the Raw SQL section is a great idea! Will try to take a look at the other updates later!

src/actions/guides/database/querying.cr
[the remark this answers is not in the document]
edwardloveall: That could work. You could list them before the examples:

> Lucky supports inequality queries like less than, greater than, and their or-equals counter parts

```crystal
# WHERE users.age <= 12
UserQuery.new.age.lte(12)

# WHERE users.updated_at < '#{3.months.ago}'
UserQuery.new.updated_at.lt(3.months.ago)
```

jwoertink:
Ok, this isn't 100% ready, but really just two more things I want to clarify.
* I want to show a larger example of a named scoped with a join like  the current docs show
* I put a note that the `has_many` method returns `Array(Task)`, but I'm pretty sure that's wrong now, but I need some clarification on what that actually is. 

Aside from those 2 things, if anyone sees anything else, please let me know.

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,526 @@ class Guides::Database::Querying < GuideAction
...
+    query for that association.
+
+    You can use this to help refine your association.
+
+    ```crystal
+    UserQuery.new.join_taks.tasks { |task_query|
+      # WHERE tasks.title = 'Clean up notes'
+      task_query.title("Clean up notes")
+    }
+    ```
```
paulcsmith: I think adding some text on what this would do in written word would help clarify too. Something like: 

> This will return all users who has a task with a title "Clean up notes"

Or  something like that

src/actions/guides/database/querying.cr
```diff
@@ -7,294 +7,526 @@ class Guides::Database::Querying < GuideAction
...
+    Return rows ordered by the `age` column in descending (or ascending) order.
+
+    `SELECT COLUMNS FROM users ORDER BY users.age DESC`
 
     ```crystal
-    UserQuery.new.none
+    UserQuery.new.age.desc_order
+    # or for asc order
+    UserQuery.new.age.asc_order
     ```
```
paulcsmith: Maybe a heading like `## Negating conditions with NOT`

That shows a few examples of using `not`. `name.not("John")` or `age.not(30)`, `name.not.ilike("Jess%")` just so it is clear it can be used in most where's and also so it shows up in search

jwoertink: I have this example `UserQuery.new.name.not.eq("Billy")`. Should this be `name.not("Billy")`?

paulcsmith: You've got it right with `not.eq` that was my bad :)

src/actions/guides/database/querying.cr
```diff
@@ -347,11 +347,22 @@ class Guides::Database::Querying < GuideAction
     When you have a model that is associated to another, your association is a method you can use
     to return those records.
 
+    ### Associations
+
+    Each association defined on your model will have a method that takes a block, and passed in the
+    query for that association.
```
edwardloveall: This is way cool. I didn't know we could do this.

jwoertink: It's a good thing these docs are coming out!

jwoertink:
Ok, like all things, this can always be improved more later, but I feel it's at a state where it offers a huge improvement over what's live now. I'm ready to merge and deploy to production if y'all are!

paulcsmith:
I totally agree! Looks good to merge whenever you are ready. 
🎉

PR #74, opened 2019-06-25
Better website grade!

### LINKED ISSUE #61: Add OG tags
For better social media sharing.

### LINKED ISSUE #68: Get our site grade up
https://website.grader.com/results/www.luckyframework.org

Right now we score a 45/100, but there's no reason we can't be at 100/100. It's actually pretty simple with just a few small things. We want to ensure that Lucky ranks well in searches. A few extra meta tags should help out.

jwoertink:
fixes #68 

This updates the site with a few things to fix our grade across several different website checkers.

* Fix some markup for W3C validation
* Adds in some SEO stuff
* Adds missing robots.txt file
* Adds some cache headers
* Minifies logo by ~59%
* Adds in open graph tags (fixes #61)
* Gives us an actual favicon

Now, granted, these scores should be taken with a grain of salt, they still provide us with better indexing in searches. One thing still missing is a sitemap. Luckily, [some dude](https://github.com/jwoertink/sitemapper) made a sitemap shard, but I left that out for now. We can add that in later to help index the blogs and guides. 

Screen Shot 2019-06-25 at 11 04 32 AM
Screen Shot 2019-06-25 at 11 07 08 AM


src/handlers/cache_control_handler.cr
```diff
@@ -0,0 +1,9 @@
+class CacheControlHandler
+  include HTTP::Handler
+
+  def call(context)
+    context.response.headers["Cache-Control"] = "public, s-maxage=#{180.days.to_i}, maxage=#{90.days.to_i}"
+    context.response.headers["Expires"] = 1.month.from_now.to_s("%a, %d %b %Y %H:%M:%S %z")
```
paulcsmith: I'm nervous about this because this sets cache control for everything, even the HTML (if I understand this correctly). I think this should only be for static assets, but HTML should not be cached (or be cached for a very short period of time). That way when we make updates people see the latest versions

src/pages/main_layout.cr
```diff
@@ -9,14 +9,14 @@ abstract class MainLayout
   # Add a `page_title` method to pages to override it. You can also remove
   # This method so every page is required to have its own page title.
   def page_title
-    "Welcome"
+    "Web framework for Crystal"
```
paulcsmith: 👍

src/utils/seo.cr
```diff
@@ -0,0 +1,3 @@
+record SEO,
+  page_title : String = "Lucky - Crystal Web Framework",
```
paulcsmith: I don't think there should be a default. That way we will know at compile time if something is missing a page title

src/utils/seo.cr
```diff
@@ -0,0 +1,3 @@
...
+  page_description : String = "Lucky is a web framework for the Crystal programming language"
```
paulcsmith: This is ok though I think since it requires a lot more thought and will require adding tons of descriptions that we don't yet have.

src/handlers/cache_control_handler.cr
jwoertink: I think if I move this after the RouteHandler, then that should work.

src/utils/seo.cr
jwoertink: For sure. We can always worry about this sort of stuff later.

src/handlers/cache_control_handler.cr
paulcsmith: Ahh yeah I think so too. Can you test on the review app to make sure the header is only present for static assets just to make sure?

jwoertink: What do you mean by "review app"? Like write the spec in this app? Or you're saying just test that in some other app?

paulcsmith: Oh I mean the Heroku review app that gets auto-deployed. Looks like it isn't deployed though so I'll make it now

jwoertink:
I looked in to the compress handler stuff. From what I can tell, and after asking on the crystal gitter, it seems the static file handler and the compress handler aren't designed to be used together. You use one or the other. We may need to make a new file to do all these things and build in to lucky 

https://github.com/luckyframework/lucky/issues/829

jwoertink:
Ok, here's the response headers from the CSS

```
HTTP/1.1 200 OK
Server: Cowboy
Date: Fri, 28 Jun 2019 18:01:59 GMT
Connection: keep-alive
Cache-Control: public, s-maxage=15552000, maxage=7776000
Expires: Sun, 28 Jul 2019 18:02:00 +0000
Etag: W/"1561744533"
Last-Modified: Fri, 28 Jun 2019 17:55:33 GMT
Content-Type: text/css
Content-Length: 332552
Via: 1.1 vegur
```

Here's the headers from the HTML doc

```
HTTP/1.1 200 OK
Server: Cowboy
Date: Fri, 28 Jun 2019 18:01:59 GMT
Connection: keep-alive
Content-Type: text/html
Transfer-Encoding: chunked
Via: 1.1 vegur
```

Looks like moving that handler worked.

paulcsmith:
Perfect! That is a cool handler!

PR #75, opened 2019-06-28
Clarify env vars

paulcsmith:


src/actions/guides/deploying/heroku.cr
```diff
@@ -156,9 +154,11 @@ class Guides::Deploying::Heroku < GuideAction
...
     ```
 
-    If you don't need to send emails, you'll need to change the config in
-    `config/email.cr` to use the `Carbon::DevAdapter`. When you open the file there
-    will be comments showing you what to do.
+    If you don't need to send emails you can set the value to `unused`.
+    
+    Alternatively, you can change the config in `config/email.cr` to use the
+    `Carbon::DevAdapter`. When you open the file there
+    will be comments showing you how to do this.
```
jwoertink: Maybe we can also link to the email docs? https://www.luckyframework.org/guides/emails/sending-emails-with-carbon

I would imagine we will beef up that page as well, but having a link here now will help later.

src/actions/guides/deploying/heroku.cr
```diff
@@ -141,12 +141,10 @@ class Guides::Deploying::Heroku < GuideAction
...
 
     ```bash
-    $ lucky gen.secret_key
-    <GENERATED_SECRET_KEY>
-    $ heroku config:set SECRET_KEY_BASE=<GENERATED_SECRET_KEY>
+    $ heroku config:set SECRET_KEY_BASE=$(lucky gen.secret_key)
     ```
 
-    ### `SEND_GRID_KEY` (optional)
+    ### `SEND_GRID_KEY` (set to "unused" if not sending emails)
```
edwardloveall: Yeah! I think this line alone will reduce a ton of questions around deployment errors

paulcsmith:
Thanks for the reviews! Pushed up some updates and will merge/deploy soon

Issue #80, opened 2019-07-06
Add generic deployment guide


paulcsmith:

mwlang:
How generic do you you want this to be?

Should I only cover components of the system that must be installed?  i.e. Crystal, yarn, etc. 

Or, should I say, write instructions and commands to issue to install a complete system onto Ubuntu 18.04 and MacOS that can likely be interpolated to other system installs?

paulcsmith:
I think at first just what components to instal and what commands to run. Then later what commands to install

But if you have them ready please share a full install on Ubuntu as we may use that in the future!

edwardloveall:
Yeah, honestly we need _any_ start we can get. If you have very specific Ubuntu instructions, those are way more helpful than nothing. Same goes for generic instructions or anywhere in between.

jwoertink:
This has been added in with https://github.com/luckyframework/website/pull/81

PR #81, opened 2019-07-09
WIP: Ubuntu deployment guide

oneiros:
A little background first:

Last year I deployed my first lucky app to production. I used capistrano to automate the deployment process. It took me some time but in the end I was quite happy with the result. I always wanted to do a write up, but never got around to it.

Fast forward several (too many) months, and discussions on gitter and #80 reminded me to finally write something. As @edwardloveall said in #80, anything could be helpful.

So here it is. Some of this is from memory. Some of it is extracted from my running production app. The app is still on lucky 0.11, so some of this might be outdated.

I will try to find some time to set up a super simple app with current lucky and test, step-by-step, if my instructions actually work.

### DIFF
--- a/src/actions/guides/deploying/ubuntu.cr
+++ b/src/actions/guides/deploying/ubuntu.cr
@@ -1,0 +1,472 @@
+class Guides::Deploying::Ubuntu < GuideAction
+  guide_route "/deploying/ubuntu"
+
+  def self.title
+    "Deploying to an Ubuntu server"
+  end
+
+  def markdown
+    <<-MD
+    ## Prerequisites 
+
+    What you will need:
+
+    * A server with Ubuntu Linux (
+      this guide assumes 18.04 but should work for
+      other versions a well)
+    * SSH access to the server, either as `root` user
+      or a user with `sudo` rights (preferred)
+    * A lucky app with an awesome name. Replace
+      `<yourapp>` in the following instructions with
+      that name.
+    * A domain name. While lucky apps can be deployed
+      in a "subdirectory" of a domain, they work best
+      when installed at the root of the domain.
+      Replace `<yourdomain>` in the following
+      instructions with that domain name.
+
+    ## Install required dependencies
+
+    * Follow the
+      [official installation instructions](https://crystal-lang.org/reference/installation/on_debian_and_ubuntu.html)
+      to install Crystal. Also install the mentioned 
+      optional (but recommended) packages.
+      (Please note that even if you do not plan to
+      compile code on your server, you still need
+      to install Crystal, because you need its runtime.)
+    * Follow the [installation instructions](https://yarnpkg.com/lang/en/docs/install/#debian-stable) to install yarn
+      (This is optional and only needed if you
+      plan to compile assets on the server.)
+    * Install PostgreSQL
+      ```bash
+      sudo apt install postgresql
+      ```
+    * Install git
+      ```bash
+      sudo apt install git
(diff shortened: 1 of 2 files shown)


src/actions/guides/deploying/ubuntu.cr
```diff
@@ -0,0 +1,472 @@
...
+  def markdown
+    <<-MD
+    ## Prerequisites 
+
+    What you will need:
+
+    * A server with Ubuntu Linux (
+      this guide assumes 18.04 but should work for
+      other versions a well)
+    * SSH access to the server, either as `root` user
```
jwoertink: Setting up a server with root access isn't really recommended, so what are your thoughts on just assuming the user with access to a `deploy` user?

I'm thinking maybe something like this.

```diff
- SSH access to the server, either as `root` user or a user with `sudo` rights (preferred)
+ SSH access to the server with `sudo` permissions.
```

src/actions/guides/deploying/ubuntu.cr
```diff
@@ -0,0 +1,472 @@
...
+      ```
+      (This is optional and only needed if you
+      plan to deploy source code from a git
+      repository.)
+
+    ## Add a deployment user
+
+    While not strictly necessary, it is highly
+    recommended to create a dedicated user to
+    deploy your app. E.g.:
```
jwoertink: Like above, I think we should just focus on the one way to do this. I'm not sold on the wording, but maybe this will give you some ideas.

```diff
-  While not strictly necessary, it is highly recommended to create a dedicated user to deploy your app. E.g.:
+  We will create a `deploy` user for deploying and managing our app.
```

src/actions/guides/deploying/ubuntu.cr
```diff
@@ -0,0 +1,472 @@
...
+    ```bash
+    crystal run tasks.cr -- db.migrate
+    ```
+
+    You could skip this part if you did not
+    add any migrations, but the task is smart
+    enough to notice this anyway. So it is
+    better to get in the habit to run it
+    every time, so not to forget it when it
+    is really needed.
```
jwoertink: I think it's ok to just run this every time, so we can probably just remove this paragraph. Or use `>` so it's a side note like `> You may skip this on a first deploy, or if there's no migrations`

jwoertink (commented):
This is a great guide! I haven't had a chance to review the whole thing just yet, but a few comments.

Since we know there's a million ways you can setup a deployment, my thought is what if we structured this in a way to remove many of the "optional" parts. Make the guide more like "Do X, then do Y", as opposed to "You can do X, or Y, but we will do Z". I think if we can make this guide smaller and more concise, it'll be easier for people to digest.

Also along those lines where you mention things like "You can use root", but since using "root" isn't really recommended for security reasons, we just remove those things.

oneiros:
> I think if we can make this guide smaller and more concise, it'll be easier for people to digest.

I think you are absolutely correct. Thank you for the feedback and sorry for taking so long to reply.

I went ahead and removed all the optional stuff. I may turn some of that into blog posts at some point, but for now it is more important for the guide to help people get to a running app as quickly as possible.

I have not yet tested the instructions in a real world scenario, but hope to get around to it in the coming weeks.

paulcsmith:
@oneiros This is freakin incredible! I think I'm gonna go ahead and merge this. We can test this out and if it needs tweaks we can make those as we move forward. Thank  you so much for this 💚

PR #85, opened 2019-07-16
add explaining Intro sentence to examples

testbird:


src/pages/main_layout.cr
```diff
@@ -109,7 +109,13 @@ abstract class MainLayout
     If this sounds interesting, [let's get started](#{Guides::GettingStarted::Installing.path}).
 
     ## What does Lucky look like?
-
+    
+    In Lucky we generate separate action classes with names that describe the route,
+    and they define the routing paths together with the response.
```
paulcsmith: This second part "and they define the routing paths" sounds similar to "that describe the route". So I think "and they define the routing paths" can be removed

paulcsmith (commented):
Just one small comment and then ready to go!

src/pages/main_layout.cr
testbird: I already wanted to hint at the generator working based on the name.

Maybe this:
> In Lucky we generate separate action classes with names that describe the route.
> The resulting file maps routing path definitions to a response block.

Though, currently the paths may still be implicit, the mapping is there.

src/pages/main_layout.cr
```diff
@@ -110,8 +110,8 @@ abstract class MainLayout
 
     ## What does Lucky look like?
     
-    In Lucky we generate separate action classes with names that describe the route,
-    and they define the routing paths together with the response.
+    In Lucky we generate separate action classes with names that describe the route.
+    The resulting file maps routing path definitions to a response block.
```
paulcsmith, line 114: Looks great!

PR #86, opened 2019-07-16
intro starting with defaults

testbird:
Info as suggested to me by https://thoughtbot.com/blog/designing-lucky-actions-routing

### DIFF
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -88,20 +88,19 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
 
     ### Path parameters
 
-    When defining an explicit path, you may mark certain parts of the path for later access as parameters.
-    Sections of the path that start with `:` will generate a method returning that param in your action.
+    When defining an explicit path, you may mark certain parts of the path, to have them available later.
+    Sections of the path that start with a `:` will generate a method that returns that param in the action.
     ```crystal
     # src/actions/users/show.cr
     class Users::Show < BrowserAction
       get "/users/:my_user_id" do
-        text "User with an id of \#{my_user_id}"
+        text "User has id \#{my_user_id}"
       end
     end
     ```
     In this case anything you pass in the part of the URL for `:my_user_id` will
     be available in the `my_user_id` method. So in this example if you visited
-    `/users/123` then the `my_user_id` would return a text response of `User with
-    an id of 123`.
+    `/users/123` then the `my_user_id` would return a text response of `User has id 123`.
 
 
     ### You can use as many parameters as you want
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -22,7 +22,7 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     Lucky automatically configures [RESTful routes (see below)](##{ANCHOR_AUTOMATICALLY_GENERATE_RESTFUL_ROUTES}).
     
     For example, an action class `Item::Show` defaults to handle `get "/item/:item_id"`, a HTTP GET request to show a specific item.
-    The route can also be defined for specific request types by using `get`, `put`, `post`, `patch`, `trace`, and `delete` macros.
+    The route can also be defined for specific request types by using the `get`, `put`, `post`, `patch`, `trace`, and `delete` macros.
 
     If you need access to a different method like `options`, you can also use the `match` macro.
 
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -16,18 +16,29 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
 
     ## Routing
 
-    Unlike many frameworks, Lucky has no separate definition file for routes.
+    Instead of separate definition files for routes and cotrollers, Lucky uses action classes.
+    This allows for solid error detection, and method and helper creation.
 
-    And Lucky can automatically infer the route path if the name of the class ends with one of the known [RESTful actions (see below)](##{ANCHOR_AUTOMATICALLY_GENERATE_RESTFUL_ROUTES}).
+    A default route path is automatically inferred from the name of the class, if it ends with a known [RESTful action (see below)](##{ANCHOR_AUTOMATICALLY_GENERATE_RESTFUL_ROUTES}).
     
-    For example, an action named `Item::Show` defaults to handle `get "/item/:item_id"`,
-    a HTTP GET request to show a specific item.
-    Executing `lucky gen.action Item::Show` will create the action with this default route.
+    For example, an `Item::Show` action will by default respond to `get "/item/:item_id"`, a HTTP GET request
+    for a specific item, and have the requested item_id available as #{:item_id}.
+
+    Let's generate an index action for showing users with
+    `lucky gen.action.browser Users::Index`
+    to see what a simple action looks like:
+
+    ```crystal
+    class Users::Index < BrowserAction
+      get "/users/:user_id" do
+        # `text` sends plain/text to the client
+        text "Rendering something in Users::Index"
+      end
+    end
+    ```
     
-    The macros `route` and `nested-route` do also still automatically imply these default route paths,
-    however, their depreciation is [discussed](https://github.com/luckyframework/lucky/issues/789). 
     
-    Explicit routes can be defined for specific request types by using the `get`, `put`, `post`, `patch`, `trace`, and `delete` macros.
+    Routes can be defined manually for specific request types by using the `get`, `put`, `post`, `patch`, `trace`, and `delete` macros.
     
     If you need access to still different methods like `options`, you can use the `match` macro.
 
@@ -41,20 +52,7 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     end
     ```
 
-    Let's generate an index action for showing users with `lucky
-    gen.action.browser Users::Index` to see what a simple action looks like:
 
-    ```crystal
-    # src/actions/users/index.cr
-    class Users::Index < BrowserAction
-      # The name of this action lets it respond to
-      # GET requests for the path "/users" by default.
-      route do
-        # `text` sends plain/text to the client
-        text "Rendering something in Users::Index"
-      end
-    end
-    ```
 
     > Note `lucky gen.action.browser` is used to create actions that should be
     shown in a browser. You can also use `lucky gen.action.api` for actions meant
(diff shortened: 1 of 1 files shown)


src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -16,11 +16,13 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
-
-    You can route to an action by using `get`, `put`, `post`, `patch`, `trace`, and `delete` macros, or
-    [Lucky can figure it out for you](##{ANCHOR_AUTOMATICALLY_GENERATE_RESTFUL_ROUTES}).
+    Unlike many frameworks, there is no separate route definition file.
+    
+    The name of an action class defines the route that triggers that action by default.
+    Lucky automatically configures [RESTful routes (see below)](##{ANCHOR_AUTOMATICALLY_GENERATE_RESTFUL_ROUTES}).
+    
+    For example, an action class `Item::Show` defaults to handle `get "/item/:item_id"`, a HTTP GET request to show a specific item.
+    The route can also be defined for specific request types by using `get`, `put`, `post`, `patch`, `trace`, and `delete` macros.
```
paulcsmith: We may actually remove this functionality https://github.com/luckyframework/lucky/issues/789

So I think we should leave this as it was since we don't want over  emphasize something that is likely to be removed

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -40,10 +42,11 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     ```crystal
     # src/actions/users/index.cr
     class Users::Index < BrowserAction
-      # GET requests to the /users path are handled by this action
+      # The name of this action lets it respond to
+      # GET requests for the path "/users" by default.
```
paulcsmith: Looks good!

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -85,24 +88,20 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
 
     ### Path parameters
 
-    Sometimes you want to name certain parts of the path and access them as parameters.
-
+    When defining an explicit path, you may mark certain parts of the path, to have them available later.
```
paulcsmith: I think the "to have them available later" could be clarified some. Maybe "you may mark certain parts of dynamic parameters that can be accessed from within the action"?

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -85,24 +88,20 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    Sections of the path that start with a `:` will generate a method that returns that param in the action.
```
paulcsmith: 👍

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -85,24 +88,20 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
     ```crystal
     # src/actions/users/show.cr
     class Users::Show < BrowserAction
       get "/users/:my_user_id" do
-        text "User with an id of \#{my_user_id}"
+        text "User has id \#{my_user_id}"
```
paulcsmith: I don't think either of these make that much sense since we didn't actually find a user. Maybe `text "user_id from path is: #{my_user_id}"`?

paulcsmith (commented):
A few comments. Thank you so much for improving the guides. This makes it soo much nicer

src/actions/guides/http_and_routing/routing_and_params.cr
testbird: Too bad, I liked that short syntax. What about adding 

```suggestion
    Lucky started out automatically configuring [RESTful routes (see below)](##{ANCHOR_AUTOMATICALLY_GENERATE_RESTFUL_ROUTES}), based on the class name.
    For example, an action class `Item::Show` defaults to handle `get "/item/:item_id"`,
    a HTTP GET request to show a specific item. However, removal of this feature
    is [discussed](https://github.com/luckyframework/lucky/issues/789). 
    
    Explicit routes can be defined for specific request types by using the `get`, `put`, `post`, `patch`, `trace`, and `delete` macros.
```

testbird:
I don't understand that "Build Errored" message.

src/actions/guides/http_and_routing/routing_and_params.cr
paulcsmith: Since this will go live right after I merge I think that may be confusing since you currently can use the `route` macro.

The plan is that the generators will still be able to infer the route name for you so it can still be fast to add new actions. The other reason for this change is that we want to allow specifying path types, which is hard/impossible/weird with the `route` macro:

```crystal
# Will make sure the id can be parsed to a User::ID and can only be used in User queries
get "/users/:user_id(User::ID)" do
end
```

I hope that kind of explains the rationale. If we find a way to bring it back though maybe we will! We're still experimenting with the ideal APIs :)

paulcsmith:
It looks like the conversations were resolved but the text is the same. Maybe you have some stuff locally that needs to be pushed?

Yeah the build just stopped working and it isn't your PR. I'll have to dig into it and figure out what's wrong with CI later :S

src/actions/guides/http_and_routing/routing_and_params.cr
testbird: could route accept a class when building
route Show => User::ID

```suggestion
    Lucky can automatically infer [RESTful routes (see below)](##{ANCHOR_AUTOMATICALLY_GENERATE_RESTFUL_ROUTES}), based on the name of the class.
    For example, an action named `Item::Show` defaults to handle `get "/item/:item_id"`,
    a HTTP GET request to show a specific item.
    Executing `lucky gen.action Item::Show` will create the action with this default route.
    The `route` and `nested-route` macros do also still automatically create these default route paths,
    however, their depreciation is [discussed](https://github.com/luckyframework/lucky/issues/789). 
    
    Explicit routes can be defined for specific request types by using the `get`, `put`, `post`, `patch`, `trace`, and `delete` macros.
```
Not sure, do you mean something like this:
```
class Users::Show < BrowserAction
  match_params_to[User::ID]
  route
    text "Now #{user_id} is ensured to be a known user."
  end
end
```

paulcsmith: But what if there are multiple params?

```crystal
class Users::Projects::Show < BrowserAction
  match_param user_id : User::ID
  match_param project_id : Project::Id

  route
    text "Now #{user_id} is ensured to be a known user."
  end
end
```

The problem here is it is much more verbose and requires inner knowledge of how Lucky is generating params. Whereas `get "/users/:user_id(User::ID)/projects/:project_id(Project::ID)"` is not much typing and makes a lot of sense just from looking at it. At least that's my opinion :P But if you have ideas please leave them on that issues so we can keep track of everything

testbird: BTW what was the reason to drop `action`?
To me, `route` lets me expect path definition stuff to be in the block, wheras the path context needs to be thought of looking at the class name, whereas looking at the code block it is all about the action.

testbird: Think: `action` determines the default routing path automatically, implicitly.

paulcsmith: People were confused by `action` almost universally. `route` clicked a bit better it seemed. But I think it is subjective and depends on what you're used to or expect 🤷‍♂️ So far I haven't heard complains about the `route` name

testbird: Yeah, I think what was missing was just an explanation to understand how in Lucky the object focus is on the "action class", instead of "routes and controller files".

When the intro properly explains it (is easy to understand for new and experienced users) I think it could have actually been better to use `action` instead of `route` to let the focus stay on the action name.

But with dropping the `route` macros as well it will be mood of course, and `get` et. al. are clearer for the long run readability as you were able to convnice me.

testbird:
OK, I've gone through the file again, not sure what the problem was with the "resolve" buttons. Do you see the changes from today? https://github.com/testbird/website/commits/patch-2

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -88,19 +90,30 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    ```
+    Here, any string from the request will be returned by the `some_user_id` method. So in this example if 
+    `/users/1-2-foobar` is requested `some_user_id` would return a text response of `Requested user id: 1-2-foobar`.
+
+    To ensure that the action is only called with an actually existing User::ID the class can be passed
+    to the helper in the path definition `get "/users/:user_id(User::ID)"`:
+    
+    ```crystal
+    class Users::Show < BrowserAction
+      get "/users/:user_id(User::ID)" do
```
paulcsmith, line 114: This part is not yet implemented (planned for future release)

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -130,17 +143,22 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
 
     ### `route`
 
-    For `route`, it will use the first part of the class name as the resource name,
+    [The macros `route` and `nested-route` do still exist, and automatically imply the default route paths,
```
paulcsmith: `nested-route` should be `nested_route`.

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -130,17 +143,22 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    however, their depreciation is [discussed](https://github.com/luckyframework/lucky/issues/789). Moving
```
paulcsmith: I believe this should be `deprecation`

paulcsmith (commented):
Just a couple small comments then ready to go!

src/actions/guides/http_and_routing/routing_and_params.cr
testbird: Um no, that sounds rather hard to me. Think `depreciated` is common for phasing out software functions.

paulcsmith: Depreciation is for finance assets https://www.investopedia.com/terms/d/depreciation.asp 

Deprecation is for software https://techterms.com/definition/deprecated so I think we should be use `deprecation`

src/actions/guides/http_and_routing/routing_and_params.cr
testbird: hope the collapsible works

src/actions/guides/http_and_routing/routing_and_params.cr
testbird: oops

src/actions/guides/http_and_routing/routing_and_params.cr
testbird: Ok, distinguished language, seems to be depend on the project.

paulcsmith (approved):
Thank you!

PR #88, opened 2019-07-17
Clarify fill_existing_with

paulcsmith:


src/actions/guides/database/migrations.cr
```diff
@@ -155,9 +157,18 @@ class Guides::Database::Migrations < GuideAction
...
+      add otp_code : String, fill_existing_with: "fake-otp-code-123"
     end
     ```
+    
+    > You can also use `fill_existing_with: :nothing` if your table is empty.
+    
+    If a static value will not work, try this:
+    
+    * If you have not yet released the app, consider using `fill_existing_with: :nothing`
+      and dropping the database `lucky db.drop` and recreating it `lucky db.create && lucky db.migrate`.
```
jwoertink:
```suggestion
      and dropping the database `lucky db.drop` and recreating it with `lucky db.create && lucky db.migrate`.
```

src/actions/guides/database/migrations.cr
```diff
@@ -155,9 +157,18 @@ class Guides::Database::Migrations < GuideAction
...
+    * Consider making the type nilable `add otp_code : String?`, then fill the values with whatever value you need.
+      Then later make it required with `make_required :otp_code`
```
jwoertink, line 171: Where would `make_required :otp_code` go? Is that a model method?

jwoertink (changes requested):

src/actions/guides/database/migrations.cr
paulcsmith: It's available in migrations. I'll add an example!

src/actions/guides/database/migrations.cr
```diff
@@ -155,9 +157,33 @@ class Guides::Database::Migrations < GuideAction
...
+    end
+    
+    # Then add values to it
+    UserQuery.new.each do |user|
+      User::BaseForm.udpate!(user, otp_code: CodeGenerator.generate)
+    end
+    
+    # Then make it non-nullable
+    make_required :otp_code
+    ```
```
paulcsmith, line 188: Example! ^

src/actions/guides/database/migrations.cr
```diff
@@ -155,9 +157,33 @@ class Guides::Database::Migrations < GuideAction
...
+      add otp_code : String?
+    end
+    
+    # Then add values to it
+    UserQuery.new.each do |user|
+      User::BaseForm.udpate!(user, otp_code: CodeGenerator.generate)
+    end
+    
+    # Then make it non-nullable
+    make_required :otp_code
```
jwoertink: Just to make this super crystal clear, can you put it all inside `def migrate`? Most the other migration examples do that as well. This is just to show that these aren't 3 code snippets used around the app somewhere, but literally all in the same method.

src/actions/guides/database/migrations.cr
jwoertink: nice! I didn't know you could do that, so that's neat!

src/actions/guides/database/migrations.cr
paulcsmith: Good point! Done!

And yeah there is `make_optional` and `make_required` :) I guess those aren't documented lol

jwoertink (approved):

PR #89, opened 2019-07-18
removed superfluous parts

testbird:
Seeing it on the live site, the word "separate" seemed somehow superfluous, and consistently talking about classes better than mixing in files in the very beginning.

### DIFF
--- a/src/pages/main_layout.cr
+++ b/src/pages/main_layout.cr
@@ -110,7 +110,7 @@ abstract class MainLayout
 
     ## What does Lucky look like?
     
-    Lucky can generate **action classes**, from class names that determine the default routing path(s).
+    Lucky can generate **action classes**, from class names that already determine usefull default routing path(s).
     The resulting classes map the routing path definition to a response block.
     
     Using classes allows to provide very [solid](https://thoughtbot.com/blog/designing-lucky-actions-routing) automatic
--- a/src/pages/main_layout.cr
+++ b/src/pages/main_layout.cr
@@ -110,8 +110,8 @@ abstract class MainLayout
 
     ## What does Lucky look like?
     
-    Lucky will generate **action classes** from class names that determine useful default routing path(s).
-    The resulting classes map the routing path definition to a response block.
+    Lucky will generate **action classes** from class names that describe the route and determine
+    the default routing path(s). The resulting classes map the routing path definition to a response block.
     
     Using separate classes allows to provide very [solid](https://thoughtbot.com/blog/designing-lucky-actions-routing) automatic
     error detection, as well as generation of routing, path, and link helpers and methods.
--- a/src/pages/main_layout.cr
+++ b/src/pages/main_layout.cr
@@ -110,11 +110,11 @@ abstract class MainLayout
 
     ## What does Lucky look like?
     
-    Lucky will generate **action classes**, from class names that already determine useful default routing path(s).
+    Lucky will generate **action classes** from class names that determine useful default routing path(s).
     The resulting classes map the routing path definition to a response block.
     
     Using separate classes allows to provide very [solid](https://thoughtbot.com/blog/designing-lucky-actions-routing) automatic
-    error detection, as well as generation of routing, path, and link methods and helpers.
+    error detection, as well as generation of routing, path, and link helpers and methods.
     
     ### JSON API
 
--- a/src/pages/main_layout.cr
+++ b/src/pages/main_layout.cr
@@ -110,11 +110,12 @@ abstract class MainLayout
 
     ## What does Lucky look like?
 
-    Lucky's code generator takes **class names** that describe a route, e.g. `Api::Users::Show`,
-    and creates the classes with corresponding routes. The classes map the routing path definition to a response block.
+    Lucky uses **action classes** for handling HTTP request routes and responses.
+    The classes map the routes and parameters it handles to a response block.
+    Lucky can generate these classes for you with 'lucky gen.action`.
 
-    Using a class per action (route+response) allows to provide very [solid](https://thoughtbot.com/blog/designing-lucky-actions-routing) automatic
-    error detection, as well as generation of routing, path, and link helpers and methods.
+    Using a class per action (route+response) allows to provide very [solid](https://thoughtbot.com/blog/designing-lucky-actions-routing)
+    automatic error detection, as well as generation of routing, path, and link helpers and methods.
 
     ### JSON API
 
--- a/src/pages/main_layout.cr
+++ b/src/pages/main_layout.cr
@@ -111,7 +111,7 @@ abstract class MainLayout
     ## What does Lucky look like?
     
     In Lucky we generate **action classes**, from class names that determine inferred default routing path(s).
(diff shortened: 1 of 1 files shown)


src/pages/main_layout.cr
```diff
@@ -110,8 +110,8 @@ abstract class MainLayout
 
     ## What does Lucky look like?
     
-    In Lucky we generate separate action classes with names that describe the route.
-    The resulting file maps routing path definitions to a response block.
+    In Lucky we generate **action classes** with names that describe the route.
+    The classes then maps a routing path definition to a response block.
```
paulcsmith: I believe `maps` -> `map` would read better

paulcsmith:
Just one tiny comment then looks good!

src/pages/main_layout.cr
testbird: Yep.
Added some more context to the first sentence as well.

jwoertink (approved):

src/pages/main_layout.cr
```diff
@@ -110,8 +110,8 @@ abstract class MainLayout
 
     ## What does Lucky look like?
     
-    In Lucky we generate separate action classes with names that describe the route.
-    The resulting file maps routing path definitions to a response block.
+    Lucky can generate **action classes**, from class names that already determine usefull default routing path(s).
```
jwoertink: haha. I actually liked it better the other way. Also, it's not that Lucky *can* generate these classes, it *does indeed* generate separate classes. 

After reading this a few times, I'm thinking the original way made a bit more sense, but maybe with a small change.

Maybe something like
```
Lucky generates separate **action classes** that determine routing paths.
The resulting classes map the routing path definition to a response block.
```

jwoertink (commented):
Saw a few changes here. Just a small comment.

src/pages/main_layout.cr
testbird: I wanted to emphasize for newcomers (like me) that one can just tell Lucky to generate an action class of name X:Y:Z, and the name will determine the default routes. I only found this feature out with the old docs after quite a while.

testbird: What about using "will"  instead of "can" ?

jwoertink: oh yeah, for sure. We want to make sure this is as clear and easy to read as possible. Wording is pretty tough since it can be interpreted differently by different people depending on where in the world you come from. I think this is close though.

This reads well for me.
```
Lucky will generate **action classes** from class names that determine useful default routing path(s).
```

paulcsmith (commented):
Thanks for submitting this. I see where you’re coming from and that we don’t make it super obvious how to do route inference but I think that’s ok since it may be removed. 

I think we should stick with the original because we are likely to remove the route inference. Also we’re not using route inference in the example which also makes the description not match the example.

src/pages/main_layout.cr
testbird: For me too.
Also added the "separate" classes in the following explanation.

testbird:
I intentionally left out to explicitly mention the `route` inference, so the intro should actually work for all cases, now and later.

(i.e. the old inferred way (the "routing path definition" in the class is the `route`'s implicit one), the new way (generator creates explicit route definition), and with the "look like" example shown in the intro (a manual route currently, which can be interpreted by the reader as generated one later)

testbird:
@paulcsmith 

Could you have another look? I think exactly as you described it, the proposed text is not making super obvious how `route` inference works. This seems ok, on the front page, and also since `route` may be removed, and...

The version at which we arrived by now should work just as well with the current `route` feature (details to  be read about in the docs) as with the generator writing out explicit routes.

It should give a better picture for reading up on `route` later than the old version, because it conveys the idea that the routes depend on the class name. And  once the generator creates explicit rules it should work even better, because reading up on the `route` docs becomes obsolete.

src/pages/main_layout.cr
```diff
@@ -110,11 +110,11 @@ abstract class MainLayout
 
     ## What does Lucky look like?
     
-    In Lucky we generate separate action classes with names that describe the route.
-    The resulting file maps routing path definitions to a response block.
+    Lucky will generate **action classes** from class names that describe the route and determine
```
paulcsmith: I think I'm a  bit confused about Lucky generating an action class.  Lucky uses the class, but does not generate one. Maybe this would be more accurate:

> Lucky uses a class for handling HTTP requests and responses. The class determines which route and parameters it handles.
>
> Using separate classes allows to provide very [solid](https://thoughtbot.com/blog/designing-lucky-actions-routing) automatic error detection, as well as generation of routing, path, and link helpers and methods.

The last sentence is unchanged. What do you think?

testbird: I think your sentence is technically ok, of course, only it could convey more what Lucky "is like".

What I had in mind with "Lucky will  generate" was hinting at the generator commands
like `lucky gen.action.browser Users::Index`, which lower the typing burden, something that always seems to come to mind when first reading about having to create separate classes.

Should we maybe explicitly mention the `lucky gen.action`?
Or like this:
> Lucky's code generator creates **action classes**  from class names that describe the route and determine

paulcsmith: Maybe rather than adding that in the description we add a second sentence. 

> Lucky can generate these classes for you with 'lucky gen.action`

That way it's super clear what we're trying to say.

Thoughts?

testbird: I seem to be missing the cool part: User provided "REST" class name => route path definitions generated (currently through `route` inference).

paulcsmith: That will likely be removed so I don't think we should make that one of the main things we talk about on the home page. Especially since the code example doesn't use it.

testbird: Idea:
> Lucky's code generator takes class names that describe a route, e.g. `Api::Users::Show`, and creates the classes with corresponding routes. The classes map the routing path definition to a response block.

paulcsmith: I understand wanting to be super clear and show all the goodies, but this isn't *quite* right yet though since it doesn't generate the corresponding route yet, it uses `route`. But since we don't want to advertise that yet since it may change or be removed. I also don't want to go too into the weeds with generators. 

To get the best of both worlds we should go with:

> Lucky uses a class for handling HTTP requests and responses. The class determines which route and parameters it handles. Lucky can generate these classes for you with 'lucky gen.action`
>
> Using separate classes allows to provide very solid automatic error detection, as well as generation of routing, path, and link helpers and methods.

This covers that each class handles requests, defines routes, and that you can use a generator to get things done quickly. That covers things that will for sure be in Lucky, even in future versions

testbird: Ok, I just hope you're still planning to move the default REST class naming routes from `route` into the generator.

paulcsmith: For now that is the plan. If we remove `route` we will make the generator generate `get "/users/:user_id"` or whatever :) so it is still fast

paulcsmith (approved):
Thanks for helping us flesh  this out!

PR #95, opened 2019-08-02
Lucky 0.16

jwoertink:
This is just a minor update ready for Lucky 0.16 showing crystal 0.30 as a dep.


jwoertink:
Not sure what's with the build error.

jwoertink:
Ok, I decided to add a blog post too just so it's out of the way. It's a small one since this release won't really have anything new.

I should also note that I put the release as Monday, but I don't know when it'll actually be released.

PR #93, opened 2019-07-30
Lucky v0.17 Guides/database updates

jwoertink:
This PR updates the database guides for the next release of Avram. 

I've broken out the updates in to separate commits to hopefully make it a little easier to read through. There's still a few more PRs to be merged in soon so I'll need to update docs on those once they are merged as well.

PR #98, opened 2019-08-07
Updates for the lucky 0.17 release

jwoertink:
https://github.com/luckyframework/website/pull/93 updates the guides for the database. This one updates the rest of the guides for 0.17 release.


src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -247,6 +243,35 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+      end
+
+      memoize def calculate_numbers
+        # This is ran only the first time it's called
+        ReportQuery.new.fetch_numbers_for_today(Time.utc)
+      end
+    end
+    ```
+
+    The `memoize` macro can be used anywhere in your application to memoize a method. However,
```
paulcsmith: This isn't quite right I think. It can be used in Actions because we include Lucky::Memoizable. It can be used elsewhere but you need to include that module first. So I think we should mention that, Other than that this looks great

paulcsmith (commented):
Just one comment :D

src/actions/guides/http_and_routing/routing_and_params.cr
jwoertink: Ah that's true. Maybe we remove it for now, but in the update then we can add examples on using it in other places. Less confusion for now. Plus we may end up updating the macro to support more things.

paulcsmith: That sounds good to me 👍

paulcsmith (approved):

paulcsmith (approved):

PR #102, opened 2019-08-11
Updating lucky 0.17 compatibility

jwoertink:
This bumps the site to use Lucky 0.17 and crystal 0.30.

One thing to note is that the original [icyleaf/markd](https://github.com/icyleaf/markd/pull/17) is waiting on a PR for crystal 0.30 support. In the meantime, we can use that fork.


paulcsmith (approved):
Omg CI is passing! Ship it!

PR #108, opened 2019-08-14
Lucky 0.17

paulcsmith:


paulcsmith (commented):
I'd like to add some examples of the new stuff. Would love some ideas if you have some

src/posts/lucky_0.17.cr
```diff
@@ -0,0 +1,41 @@
...
+    ### What's new?
+
+    You can see a [full list of
+    changes](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md#changes-in-v017),
+    but here are some highlights.
+
+    [Avram](https://github.com/luckyframework/avram) has seen the most changes
+    and improvements.
+
+    * Can now use any column as the primary key. Can also use Int64, UUID or any custom type you implement.
```
jwoertink, line 29: > Can also use Int64, UUID or any custom type you implement.

You can also use `Int16` and `Int32`. Maybe we can link to the [supported typed](https://www.luckyframework.org/guides/database/migrations#primary-keys) ?

src/posts/lucky_0.17.cr
```diff
@@ -0,0 +1,41 @@
...
+    * Added ability to customize default columns (primary key, timestamps, etc.).
+    * Lots of improvements to how we save and query records.
+
```
jwoertink, line 32: Maybe instead of bullet points, each one is a sub heading, then an example could be added?

```
### Better Primary key support

We have added a `primary_key` method which allows you to name the primary key anything you want, as well as define the type as any [supported type](link). You can even customize the primary key type with whatever you implement.

block_of_code snippet


### JSON and Array support

### Multi DB support

Need to use more than 1 database with your app? We got you covered..

### Polymorphic associations

### New Db tasks

migration status, verify connection, rollback_to....
```

paulcsmith: I like that idea a lot! And I forgot about a lot of that so thank you

src/posts/lucky_0.17.cr
```diff
@@ -0,0 +1,41 @@
...
+  def summary
+    <<-TEXT
+    Lucky v0.17 is a huge release with many enhancements to Avram, Lucky's
+    database library for Postgres.
+    TEXT
+  end
+
+  def content
+    <<-MD
+    Lucky v0.17.0 is now out and has support for the newest Crystal (v0.30.0)!
```
edwardloveall, line 18: This is the most pedantic thing I've ever pointed out but I think either this should not have the trailing `.0` or we should include it in the `summary`

```suggestion
    Lucky v0.17 is now out and has support for the newest Crystal (v0.30.0)!
```

src/posts/lucky_0.17.cr
edwardloveall: Might be good to say something about the default type and also the wording is confusing.

```suggestion
    * Can now use any column as the primary key. The default has changed to `Int64` (from `Int32`). We have a few [supported types](https://www.luckyframework.org/guides/database/migrations#primary-keys) out of the box, but you can also create your own custom types.
```
Not sure where the docs are for creating your own though...

src/posts/lucky_0.17.cr
```diff
@@ -0,0 +1,41 @@
...
+
+    You can see a [full list of
+    changes](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md#changes-in-v017),
+    but here are some highlights.
+
+    [Avram](https://github.com/luckyframework/avram) has seen the most changes
+    and improvements.
+
+    * Can now use any column as the primary key. Can also use Int64, UUID or any custom type you implement.
+    * Added ability to customize default columns (primary key, timestamps, etc.).
```
edwardloveall, line 30:
```suggestion
    * Added ability to customize default columns (e.g. never include primary key, always include timestamps, etc.).
```

src/posts/lucky_0.17.cr
edwardloveall: Ooo we could link to this: https://dev.to/jwoertink/using-custom-ids-in-lucky-4ibh

jwoertink:
Closing this out in favor of https://github.com/luckyframework/website/pull/115

PR #110, opened 2019-08-14
moved guide redirects to a fallback to catch nested route changes

jwoertink:
Going to `/guides/database/querying` was throwing a 404 because it would never match this redirect path. Moving this to a `fallback` route now lets us handle all redirects regardless of their nesting. 

`request.path` will look like `"/guides/overview"` or `"/guides/database/querying"` so we just remove the guides part, and do the lookup.


paulcsmith (commented):
I left a note in Gitter but I think we should move all the redirects to a LegacyRedirectHandler so we can put all the redirects in one spot. That way we can also use the fallback for something else if we need it later

jwoertink:
Cool. Sounds good.

jwoertink:
not sure why that build failed.

paulcsmith (approved):
Perfect!

Issue #29, opened 2019-05-17
Fix "edit on GitHub" link when deployed to prod


paulcsmith:

paulcsmith:
For now going to move this to "Post launch" and fix it later

paulcsmith:
For more info, here is where the link was removed: https://github.com/luckyframework/website/commit/1b81beabd21d1a6d045621dfd9f1035146ba7e11#diff-e0a45fb5de2e681e0e5cc2a80be52e01

This will be difficult to fix w/o having Heroku deploy access though since the issue appears in prod because of the build slug path being different at compile time than runtime.

PR #114, opened 2019-08-19
Fix docs contribute link

westonganger:
Also would be nice to have a docs contribute link on every page of the docs.

### DIFF
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -60,7 +60,7 @@ class Guides::GettingStarted::Installing < GuideAction
       it to `/usr/local/bin` and it should work: `mv lucky /usr/local/bin`.
 
     If you needed different steps, please help contribute to this section by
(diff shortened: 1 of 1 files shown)


jwoertink (approved):

jwoertink:
Good catch @westonganger. Thanks.

Not sure what you mean by "a docs contribute link" though. When you say "docs", do you mean "guides"? Because this link you updated is a contribute link on the guides, but I'm probably misunderstanding.

westonganger:
Yes, this is a link to contribute to the guides. However this is the only page which contains this link. 

Would be nice to have a "Contribute to this Guide" button on every single guide page similar to other documentation sites.

jwoertink:
Ah. Yeah, we originally did on the old site. https://github.com/luckyframework/website/issues/29 This issue is related to fixing that. 

Looks like this page was just missed when updating to the new system. If you'd like to give that PR a shot, we can for sure take a look at it!

PR #122, opened 2019-09-02
Fix typo

vlazar:


paulcsmith (commented):
Hi @vlazar!

It looks like the previous change was correct. On line 100 it has `plain_text "Requested user id: \#{some_user_id}"` so the response would be `Request used id: 123-foo`.

I'm going to go ahead and close this, but if you have a suggestion for clarify this, please let me know

vlazar:
@paulcsmith Yes, the response would be `Request used id: 123-foo` but `some_user_id` would return `123-foo`, not `Request used id: 123-foo` as mentioned currently.

paulcsmith (commented):
Oh I see! I misunderstood. Nice catch :)

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -103,7 +103,7 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     ```
 
     Here, the string from the request path will be returned by the `some_user_id` method. So in this example if 
-    `/users/123-foo` is requested `some_user_id` would return a text response of `Requested user id: 123-foo`.
+    `/users/123-foo` is requested `some_user_id` would return `123-foo`.
```
jwoertink: Can you also add that the action would return a text response of `Requested user id: 123-foo`? I'm thinking something like 

> `some_user_id` would return `123-foo`, and the action would return a text response of `Requested user id: 123-foo`

paulcsmith: Good idea. I think that would help clarify things

vlazar: Updated.

I also wonder if

> would return `123-foo`

Should be

> `some_user_id` would return `"123-foo"`

jwoertink (approved):
yup, I like it. Thanks!

paulcsmith:
Nice! :D

PR #130, opened 2019-09-14
Fix issues and rework HTML guides

paulcsmith:


src/actions/guides/frontend/rendering_html.cr
```diff
@@ -399,21 +400,18 @@ class Guides::Frontend::RenderingHtml < GuideAction
...
-    layouts call `shared_layout_head`. This method is defined in
-    `Shared::LayoutHead`. If you look in the `src/components/shared/layout_head`
-    file you'll see that it defines and calls a `page_title` method.
-
-    This means that you can customize the page title in your page by writing a
-    `page_title` method:
+    Let's start with a quick example. In newly generated Lucky projects
+    you'll see that `MainLayout` defines a `page_title` method that is passed to
+    the `Shared::LayoutHeader` component. Any page that inherits from `MainLayout`
+    can write a new `page_title` method to override the one in `MainLayout`.
```
paulcsmith, line 406: We use a component now. 

Also we should just focus on the `page_title` method since that's what this section is about. Using methods to render stuff in layouts

jwoertink (approved):
Looks good to me!

paulcsmith:
Thanks for  the review!

PR #143, opened 2019-09-19
Revamp auth guides

### LINKED ISSUE #125: Add guide for JWT auth
https://github.com/luckyframework/lucky_cli/pull/395/files

## Sign Up

```
curl localhost:5000/api/sign_ups \
  -X POST \
  -d "user:email=foo@example.com" \ 
  -d "user:password=password" \
  -d "user:password_confirmation=password"
```

Which returns: `{"token":"some-long-token"}`

## Sign In

```
curl localhost:5000/api/sign_ins \
  -X POST \
  -d "user:email=foo@example.com" \ 
  -d "user:password=password" \
```

Which returns: `{"token":"some-long-token"}` or validation errors

## View Me

```
curl localhost:5000/api/me \
  -H "Authorization: Bearer {{some-long-token}}"
```

Which returns: `{"email":"foo@example.com"}`

paulcsmith:
Huge update to auth guides, and adding a  guide on API auth

Closes #125

### DIFF
--- a/src/actions/guides/authentication/api_authentication.cr
+++ b/src/actions/guides/authentication/api_authentication.cr
@@ -1,0 +1,161 @@
+class Guides::Authentication::Api < GuideAction
+  guide_route "/authentication/api"
+
+  def self.title
+    "API Authentication"
+  end
+
+  def markdown
+    <<-MD
+    ## Authenticating a request
+
+    By default all ApiActions will require an auth token because the
+    `ApiAction` in `src/actions/api_action.cr` includes the
+    `Api::Auth::RequireAuthToken`. That means that any actions inheriting
+    from it will require an auth token or the action will return a 401
+    unauthorized error.
+
+    We'll go over how to make an action accessible by unauthenticated requests
+    later in the guide.
+
+    ### Sending an authentication token
+
+    You can get an auth token by signing up/in a user, but for this example
+    we'll assume our token is `"fake-token"`.
+
+    There are 2 ways to send this token.
+
+    * With an `Authorization: Bearer` header
+    * With an `auth_token` param in the query params or in the body.
+
+    ### Examples using curl
+
+    Using an `Authorization` header:
+
+    ```bash
+    curl localhost:5000/api/some_endpoint \
+      -H "Authorization: Bearer fake-token"
+    ```
+
+    Using an `auth_token` query param:
+
+    ```bash
+    curl localhost:5000/api/some_endpoint?auth_token=fake-token
+    ```
+
+    Using a body param:
+
+    ```bash
+    curl localhost:5000/api/some_endpoint \
+      -d "auth_token=fake-token"
+    ```
+
+    > You could also include the token in JSON params: `{"auth_token":"fake-token"}`
+
+    ## Allowing unauthenticated requests
+
+    For actions that do not require an authentication token, include the
+    `Api::Auth::SkipRequireAuthToken` mixin defined in
+    `src/actions/mixins/api/auth/skip_require_auth_token.cr`.
+
+    For example:
+
+    ```crystal
+    class Api::PublicComments::Index < ApiAction
+      include Api::Auth::SkipRequireAuthToken
+    end
+    ```
+
+    ### Allowing unauthenticated requests by default
+
+    If most of your actions do not require an auth token you can remove
+    `Api::Auth::RequireAuthToken` from `ApiAction`.
+
+    Then in your actions that *do* require authentication, add
+    `Api::Auth::RequireAuthToken`:
+
+    ```crystal
+    class Api::SecretProjects::Index < ApiAction
+      include Api::Auth::RequireAuthToken
+    end
+    ```
+
+    ## Sign up
+
+    The `Api::SignUps::Create` action in `src/actions/api/sign_ups/create.cr`
+    handles sign ups. It requires an email, password, and password_confirmation
+    and returns a token for authenticating the user.
+
+    ### Curl example
+
+    ```bash
+    curl localhost:5000/api/sign_ups \
+      -X POST \
+      -d "user:email=person@example.com" \
+      -d "user:password=password" \
+      -d "user:password_confirmation=password"
+    ```
+
+    Which returns a token like:
+
+    ```json
+    {"token":"some-long-token"}
+    ```
+
+    ### Customizing sign up
+
+    To add additional fields, change validations, or customize sign up,
+    check out the `SignUpUser` operation in `src/operations/sign_up_user.cr`.
+
+    ## Sign in (generate an auth token)
+
+    The `Api::SignIns::Create` action in `src/actions/api/sign_ins/create.cr`
+    handles creating a new token. It requires an email and password
+    and returns a token for authenticating the user.
+
+    ### Curl example
+
+    ```bash
+    curl localhost:5000/api/sign_ins \
+      -X POST \
+      -d "user:email=person@example.com" \
+      -d "user:password=password"
+    ```
+
+    Which returns a token like:
+
+    ```json
+    {"token":"some-long-token"}
+    ```
+
+    ### Customizing sign in
+
+    To add additional attributes, change validations, or customize sign in,
+    check out the `SignInUser` operation in `src/operations/sign_in_user.cr`.
+
+    ## Show current user
+
+    The `Api::Me::Show` in `src/actions/api/me/show.cr` returns the user
+    for the given token. Use the token returned by sign in or sign in the
+    `Authorization` header or `auth_token` param.
+
+    ### Using curl to sign up a user
+
+    ```bash
+    curl localhost:5000/api/me \
+      -H "Authorization: Bearer {{token-returned-from-sign-in/up}}"
+    ```
+
+    Returns:
+
+    ```json
+    {"email":"person@example.com"}
+    ```
+
+    ### Customize response
+
+    You can add or modify the JSON response by changing the `UserSerializer`
+    in `src/serializers/user_serializer.cr`.
+    MD
+  end
+end
--- a/src/actions/guides/authentication/api_authentication.cr
+++ b/src/actions/guides/authentication/api_authentication.cr
@@ -1,4 +1,5 @@
 class Guides::Authentication::Api < GuideAction
+  ANCHOR_OPTIONAL_TOKEN = "perma-optional-token"
   guide_route "/authentication/api"
 
   def self.title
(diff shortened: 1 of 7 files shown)


src/actions/guides/authentication/api_authentication.cr
```diff
@@ -0,0 +1,162 @@
...
+    ```crystal
+    class Api::SecretProjects::Index < ApiAction
+      include Api::Auth::RequireAuthToken
+    end
+    ```
+
+    ## Sign up
+
+    The `Api::SignUps::Create` action in `src/actions/api/sign_ups/create.cr`
+    handles sign ups. It requires an email, password, and password_confirmation
```
jwoertink:
```suggestion
    handles user sign ups. It requires an email, password, and password_confirmation
```

src/actions/guides/authentication/browser_authentication.cr
```diff
@@ -0,0 +1,251 @@
...
+
+    ## Requiring sign in
+
+    By default actions that inherit from `BrowserAction`
+    (`src/actions/browser_action.cr`) will require sign in because
+    `BrowserAction` includes the `Auth::RequireSignIn` module
+    (`src/actions/mixins/auth/require_sign_in.cr`). This is a great way to
+    make sure that your actions are protected by default.
+
+    Typically pages that require sign in will use pages that inherit from the
```
jwoertink: This reads a little weird since all pages inherit from MainLayout anyway. Not sure of a better wording at the moment though. I'll come back to this if my brain wakes up lol

src/actions/guides/authentication/browser_authentication.cr
```diff
@@ -0,0 +1,251 @@
...
+    ### Create a new layout
+
+    This is usually the best option because you can keep the `MainLayout` for
+    signed in users, and a new layout for things like marketing pages, and
+    other public pages.
+
+    For example, if you want to add some marketing pages you could create a
+    `MarketingLayout` that either does not need a signed in user, or accepts
+    signed in users and guests.
+
```
jwoertink, line 115: Love this. It really helps to make it clear how to use all this.

jwoertink (commented):
Looks good to me. Just a few small comments.

src/actions/guides/authentication/api_authentication.cr
paulcsmith: I dig it

src/actions/guides/authentication/browser_authentication.cr
paulcsmith: Yeah this is  confusing. I'll try rewording it

paulcsmith (commented):
Thanks for the review!

src/actions/guides/authentication/api_authentication.cr
```diff
@@ -0,0 +1,162 @@
...
+    <<-MD
+    ## Authenticating a request
+
+    By default all actions inherited from `ApiAction` will require an auth
+    token because the `ApiAction` in `src/actions/api_action.cr` includes the
+    `Api::Auth::RequireAuthToken`. This modules requires an auth token or the
+    action will return a 401 unauthorized error.
+
+    We'll go over how to make an action accessible by unauthenticated requests
+    later in the guide.
```
edwardloveall, line 19: Can we link to where "later" is?

src/actions/guides/authentication/api_authentication.cr
```diff
@@ -0,0 +1,162 @@
...
+      -X POST \
+      -d "user:email=person@example.com" \
+      -d "user:password=password" \
+      -d "user:password_confirmation=password"
+    ```
+
+    Which returns a token like:
+
+    ```json
+    {"token":"some-long-token"}
```
edwardloveall, line 103:
```suggestion
    {"token":"new-valid-token"}
```

I also considered `fake-token` like you have elsewhere, but that might imply that you're always returning the same token if someone signs up.

src/actions/guides/authentication/api_authentication.cr
```diff
@@ -0,0 +1,162 @@
...
+    ### Customizing sign in
+
+    To add additional attributes, change validations, or customize sign in,
+    check out the `SignInUser` operation in `src/operations/sign_in_user.cr`.
+
+    ## Show current user
+
+    The `Api::Me::Show` in `src/actions/api/me/show.cr` returns the user
+    for the given token. Use the token returned by sign in or sign in the
+    `Authorization` header or `auth_token` param.
```
edwardloveall: I'm not sure this sentence is complete? First, it's a bit hard for me to parse because `sign in` was used twice so close together. Second, did you mean to say something like:

> ... or sign in **with** the `Authorization` header ...

Third, joining these two concepts with `or` feels like a non-sequitur:

* Use X from Y: Use the token returned by sign in
* Use X to do Y: sign in the `Authorization` header or `auth_token` param.

src/actions/guides/authentication/browser_authentication.cr
```diff
@@ -0,0 +1,251 @@
...
+      #
+      # needs current_user : User?
+    end
+    ```
+
+    Some other ideas for naming layouts:
+
+    * `PublicLayout`
+    * `GuestLayout`
+    * `HelpLayout`
```
edwardloveall: Small thing, but what do you think about support instead of help? Help feels a little more ambiguous to me, maybe?

```suggestion
    * `SupportLayout`
```

src/actions/guides/authentication/intro.cr
```diff
@@ -0,0 +1,153 @@
...
+    and [APIs](#{Guides::Authentication::Api.path}), but here
+    are some quick examples of working with the signed in user.
+
+    ### Get the signed in user
+
+    To get the currently signed in user, call `current_user` in your actions.
+    If your action requires sign in, `current_user` will always return a `User`
+    and never `nil`. If sign in is
+    [optional](#{Guides::Authentication::Browser.path(anchor: Guides::Authentication::Browser::ANCHOR_ALLOW_GUESTS)})
+    then  `current_user` may be `nil`.
```
edwardloveall:
```suggestion
    then `current_user` may be `nil`.
```

src/actions/guides/authentication/intro.cr
```diff
@@ -0,0 +1,153 @@
...
+    In HTML pages using the `MainLayout` you can get the signed in user with
+    `@current_user`.
+
+    ### When current user might be nil
+
+    If the current user might be `nil` by
+    [allowing guests](#{Guides::Authentication::Browser.path(anchor: Guides::Authentication::Browser::ANCHOR_ALLOW_GUESTS)})
+    or [skipping the auth token](#{Guides::Authentication::Api.path(anchor: Guides::Authentication::Api::ANCHOR_OPTIONAL_TOKEN)})
+    you need to use a conditional before using `current_user`
+
```
edwardloveall, line 46: Might be worth highlighting that this is because of crystal's `if var` functionality:

> This works because Crystal will ensure a variable isn't nil when you use it as the conditional in an `if` statement. Read more about this [here](https://crystal-lang.org/reference/syntax_and_semantics/if_var.html)

src/actions/guides/authentication/intro.cr
```diff
@@ -0,0 +1,153 @@
...
+        SaveArticle.create!(params, author_id: current_user.id) do |op, article|
+          # ...
+        end
+      end
+    end
+    ```
+
+    ### Filter records by current_user
+
+    Let's say we only want to return articles for the signed in user. We
```
edwardloveall: "Let's say" can start to sound a bit strange when used multiple times with no preface or followup.

```suggestion
    If we only want to return articles for the signed in user. We
```

src/actions/guides/authentication/api_authentication.cr
paulcsmith: Yes I missed various words here that made this confusing. Nice catch :)

src/actions/guides/authentication/intro.cr
paulcsmith: Good idea!

src/actions/guides/authentication/api_authentication.cr
paulcsmith: I think I'll stick with the original  here so that it makes sense in the other sections as well.  But  we can change it  later if people get  confused

src/actions/guides/authentication/intro.cr
paulcsmith: Yes! Good idea

src/actions/guides/authentication/browser_authentication.cr
```diff
@@ -0,0 +1,253 @@
...
+
+    By default actions that inherit from `BrowserAction`
+    (`src/actions/browser_action.cr`) will require sign in because
+    `BrowserAction` includes the `Auth::RequireSignIn` module
+    (`src/actions/mixins/auth/require_sign_in.cr`). This is a great way to
+    make sure that your actions are protected by default.
+
+    If your action requires sign in and you are rendering HTML, you will
+    likely make the page you are rendering inherit from `MainLayout`. This is
+    because `MainLayout` requires that a `current_user` is available.
```
paulcsmith, line 31: @jwoertink Is this better?

src/actions/guides/authentication/api_authentication.cr
```diff
@@ -0,0 +1,162 @@
...
+    ### Customizing sign in
+
+    To add additional attributes, change validations, or customize sign in,
+    check out the `SignInUser` operation in `src/operations/sign_in_user.cr`.
+
+    ## Show current user
+
+    The `Api::Me::Show` in `src/actions/api/me/show.cr` returns the user for
+    the given token. Use the token returned by sign in or sign up in either
+    the `Authorization` header or in an `auth_token` param.
```
paulcsmith, line 141: @edwardloveall does this make sense now?

paulcsmith (commented):
Thanks for the feedback. I think I addressed e everything but  would love a second set of eyes on the parts I rewrote to see if  it is any clearer

src/actions/guides/authentication/browser_authentication.cr
jwoertink: Looks good to me!

src/actions/guides/authentication/api_authentication.cr
edwardloveall: Yes! Nicely worded.

paulcsmith:
Thanks for the reviews! Merging and will fix CI once we have a the Crystal 0.31 fixes in :)

PR #150, opened 2019-09-23
Reference LuckyFlow dependencies

### LINKED ISSUE #100: Link to chromedriver setup
If you want to test with Flow, you need to go through these guides https://www.luckyframework.org/guides/frontend/testing#setup. But when just getting things setup initially, you may not have chrome driver installed. So when you go through the [deps guide](https://www.luckyframework.org/guides/getting-started/installing#install-required-dependencies) you don't know about installing chrome driver until you may have already hit an error. 

We should link to those guides as a note:

> If you will be testing your front-end, you will need a few additional dependencies [link to that guide]

or whatever.

xuanxu:
This PR adds a reference to the additional dependencies needed to run the frontend tests to the _Installing_ guide.
Fixes #100

### DIFF
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -20,6 +20,9 @@ class Guides::GettingStarted::Installing < GuideAction
     * [Crystal v0.30](https://crystal-lang.org/reference/installation/)
     * Postgres ([macOS](https://postgresapp.com)/[Others](https://wiki.postgresql.org/wiki/Detailed_installation_guides))
 
+    You will need additional dependencies if you want to test your frontend using LuckyFlow,
+    see the [_Testing HTML and Interactivity_ guide](#{Guides::Frontend::Testing.path}) for details.
+
     ### Debian and Fedora dependencies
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/getting-started/installing.cr
```diff
@@ -20,6 +20,9 @@ class Guides::GettingStarted::Installing < GuideAction
     * [Crystal v0.30](https://crystal-lang.org/reference/installation/)
     * Postgres ([macOS](https://postgresapp.com)/[Others](https://wiki.postgresql.org/wiki/Detailed_installation_guides))
 
+    You will need additional dependencies if you want to test your frontend using LuckyFlow,
+    see the [_Testing HTML and Interactivity_ guide](#{Guides::Frontend::Testing.path}) for details.
```
```suggestion
    see the [Testing HTML and Interactivity](#{Guides::Frontend::Testing.path}) guide for details.
```
Just for consistency, lets take off the underscore. I don't think we have any other links with italics.

jwoertink (dismissed):
Awesome! Thanks for taking care of this. Just 1 small comment and it's good to go.

src/actions/guides/getting-started/installing.cr
xuanxu: :+1: Done!

paulcsmith (approved):
Thank you! :D

PR #160, opened 2019-10-02
Update for Lucky 0.18 and Crystal 0.31.1

paulcsmith:


src/actions/guides/getting-started/installing.cr
```diff
@@ -6,22 +6,36 @@ class Guides::GettingStarted::Installing < GuideAction
...
   end
 
-  def markdown
+  def markdown : String
     <<-MD
     #{permalink(ANCHOR_INSTALL_REQUIRED_DEPENDENCIES)}
     ## Install Required Dependencies
 
-    To get Lucky, you need to install these first.
+    ### Crystal v0.31.1
```
paulcsmith, line 14: Gave each dep a subheading to make it more scannable. Would love thoughts on this

paulcsmith:
Changes deployed here: https://lucky-website-pr-160.herokuapp.com/guides/getting-started/installing

src/actions/errors/show.cr
```diff
@@ -1,52 +1,60 @@
...
   end
 
-  # This is the catch all method that renders unhandled exceptions
-  def handle_error(error : Exception)
-    Lucky.logger.error(unhandled_error: error.inspect_with_backtrace)
-
-    if Lucky::ErrorHandler.settings.show_debug_output
-      # In development and test, render a debug page
-      render_detailed_exception_page(error)
+  # Always keep this below other 'render' methods or it may override your
```
edwardloveall, line 28: Interesting I would have thought it would override if it were below other render methods, i.e. the last defined version of a method overrides all previous definitions.

src/actions/guides/getting-started/installing.cr
edwardloveall: Yeah love it. Good headings help guide you through the doc. These are good.

src/serializers/error_serializer.cr
```diff
@@ -0,0 +1,14 @@
+# This is the default error serializer generated by Lucky.
+# Feel free to customize it in any way you like.
+class ErrorSerializer < BaseSerializer
+  def initialize(
+    @message : String,
+    @details : String? = nil,
+    @param : String? = nil # If there was a problem with a specific param
```
edwardloveall: This comment didn't make sense to me for whatever reason until I read it a few times. What do you think about:

```suggestion
    @param : String? = nil # so you can track which param (if any) caused the problem
```

src/actions/guides/getting-started/installing.cr
```diff
@@ -6,22 +6,36 @@ class Guides::GettingStarted::Installing < GuideAction
...
-    * Install one of these process managers: [Overmind (recommended)](https://github.com/DarthSim/overmind#installation),
-      [Heroku CLI (great if you plan to use Heroku to deploy)](https://devcenter.heroku.com/articles/heroku-cli#download-and-install),
-      [forego](https://github.com/ddollar/forego#installation),
-      or [foreman](https://github.com/ddollar/foreman#installation).
-    * [Crystal v0.30](https://crystal-lang.org/reference/installation/)
-    * Postgres ([macOS](https://postgresapp.com)/[Others](https://wiki.postgresql.org/wiki/Detailed_installation_guides))
+    **We recommend using a version manager** to make sure the correct
+    version of Crystal is used with Lucky.
+    Try [crenv](https://github.com/pine/crenv) (recommended) or
+    [asdf-crystal](https://github.com/marciogm/asdf-crystal) (great if you use already [asdf](https://github.com/asdf-vm/asdf))
```
```suggestion
    [asdf-crystal](https://github.com/marciogm/asdf-crystal) (great if you already use [asdf](https://github.com/asdf-vm/asdf))
```

src/actions/guides/getting-started/installing.cr
```diff
@@ -6,22 +6,36 @@ class Guides::GettingStarted::Installing < GuideAction
...
 
-    You will need additional dependencies if you want to test your frontend using LuckyFlow,
-    see the [Testing HTML and Interactivity](#{Guides::Frontend::Testing.path}) guide for details.
+    Alternatively you could [install Crystal without a version manager](https://crystal-lang.org/reference/installation/).
+
+    ### Process manager
+
+    Lucky uses a process manager to watch assets and start the server.
```
```suggestion
    Lucky uses a process manager to watch assets and start the server in development.
```

src/serializers/error_serializer.cr
jwoertink: This is the comment that's generated in apps by default https://github.com/luckyframework/lucky_cli/blob/master/src/web_app_skeleton/src/serializers/error_serializer.cr#L7

I'm all down for clarity, but if we do decide to change it, then we should make the same change in the CLI.

jwoertink (approved):
sweet! Love it. I left a few suggestion, otherwise it's all good for me.

src/actions/errors/show.cr
paulcsmith: Actually you might be right. Going to open an issue to investigate this. https://github.com/luckyframework/lucky_cli/issues/422

src/serializers/error_serializer.cr
paulcsmith: I like the change and I'll change it in CLI too. Thanks both for pointing those things out

paulcsmith (commented):
Thank you for the feedback!

PR #165, opened 2019-10-05
Guides/query updates

jwoertink:
This updates the query guides with the new null sorting for orders and the new between query method.


paulcsmith (approved):
Looks good to merge and deploy!!

PR #166, opened 2019-10-05
Fix typos

vlazar:
In addition to some typos I've also updated an outdated list of built-in tasks.

### DIFF
--- a/src/actions/guides/database/models.cr
+++ b/src/actions/guides/database/models.cr
@@ -69,7 +69,7 @@ class Guides::Database::Models < GuideAction
 
     ```crystal
     class CompletedProject < BaseModel
-      # Will use :completed_projects as the tbale name
+      # Will use :completed_projects as the table name
       table do
       end
     end
@@ -177,12 +177,12 @@ class Guides::Database::Models < GuideAction
     * `Int64` - `bigint` column type.
     * `Float64` - `numeric` column type.
     * `Bool` - `boolean` column type.
-    * `Time` - `timestamp with time zone` (`timestampz`) column type.
+    * `Time` - `timestamp with time zone` (`timestamptz`) column type.
     * `UUID` - `uuid` column type.
     * `JSON::Any` - `jsonb` column type.
     * `Array(T)` - `[]` column type where `T` is any other supported type.
 
-    Any of your columns can also define "nillable" types by adding Crystal `Nil` Union `?`.
+    Any of your columns can also define "nilable" types by adding Crystal `Nil` Union `?`.
     This is if your column allows for a `NULL` value. (e.g. `column age : Int32?` allows an
     `int` or `NULL` value).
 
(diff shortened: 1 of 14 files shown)


src/actions/guides/database/models.cr
```diff
@@ -177,12 +177,12 @@ class Guides::Database::Models < GuideAction
     * `Int64` - `bigint` column type.
     * `Float64` - `numeric` column type.
     * `Bool` - `boolean` column type.
-    * `Time` - `timestamp with time zone` (`timestampz`) column type.
+    * `Time` - timestamp with time zone (`timestamptz`) column type.
```
jwoertink:
```suggestion
    * `Time` - `timestamp with time zone` (`timestamptz`) column type.
```

The backticks here are actually correct. It's what's shown if you did `\d comments` in a `psql`.

src/actions/guides/database/models.cr
```diff
@@ -367,11 +367,11 @@ class Guides::Database::Models < GuideAction
 
     The `Comment` model now has a `commentable` method which could return a `photo` object
     or a `video` object depending on which was associated.
-    <!-- go go polymorphin power rangers -->
+    <!-- go go polymorphic power rangers -->
```
jwoertink: 😂  hahahah! I actually spelled it with `morphin` on purpose because they were "mighty morphin power rangers".  It was fun, but we can just remove this line now 🤣

src/actions/guides/frontend/rendering_html.cr
```diff
@@ -140,11 +140,11 @@ class Guides::Frontend::RenderingHtml < GuideAction
     end
     ```
 
-    If you need to pass in data attributes, or any arbitrary attributes for use in SPAs (i.e. ng-app, v-bind:click, etc...), you can also use a string.
+    If you need to pass in data attributes, or any arbitrary attributes for use in SPAs (i.e. ng-app, v-on:click, etc...), you can also use a string.
 
     ```crystal
     def content
-      div(ng_model: "something", data_action: "someAction", "v-bind:click": "update")
+      div(ng_model: "something", data_action: "someAction", "v-on:click": "update")
```
jwoertink: Ha! Good catch. It's been a long time since I used angular 😄

jwoertink (changes requested):
Awesome! Thanks for catching these. I'm sure a `git blame` would point to me in most of these cases 😅  Just a few small things, and then it's good to go!

src/actions/guides/database/models.cr
vlazar: TIL. Thanks!

src/actions/guides/database/models.cr
vlazar: Ah, I was suspecting there was something to it but couldn't figure it out. Reverted this one.

src/actions/guides/frontend/rendering_html.cr
vlazar: The `v-on:click` and `v-bind` is actually Vue.js syntax. I've changed to `v-on` since `Vue.js was mentioned in other place with `v-bind:click`. I think in Angular 1 there was `ng-bind` and I think `ng-click`. Updated to Angular attrs here and left `v-on:click` in another comment mentioning Vue.js

Or maybe update this to Vue.js too since Angular 1 is kind of old, so why mention in leading edge framework? 😄

jwoertink: I probably confused the two 😅 It doesn't really matter which framework just as long as it shows that you can do custom attributes this way. I think your change here is fine 😄

jwoertink (approved):

PR #168, opened 2019-10-05
update routing and params example to user Auth:AllowGuests

bdtomlin:
The example was still including the Auth::SkipRequireSignIn module that no longer exists. I replaced the example with the current exact code that is generated with lucky init, since it has improved comments as well.


paulcsmith (approved):
Thank you!

PR #170, opened 2019-10-06
Add RSS feed to blog

wezm:
I tried to subscribe to the blog in my RSS reader but there was no feed. So this PR adds one.

Passes validation at https://validator.w3.org/feed/

### DIFF
--- a/src/actions/blog/feed.cr
+++ b/src/actions/blog/feed.cr
@@ -1,6 +1,6 @@
 class Blog::Feed < BrowserAction
   get "/blog/feed.xml" do
-    posts = PostQuery.new.all[0...10]
+    posts = PostQuery.new.all.first(10)
     xml render_feed(posts)
   end
 
--- a/src/actions/blog/feed.cr
+++ b/src/actions/blog/feed.cr
@@ -1,31 +1,6 @@
 class Blog::Feed < BrowserAction
   get "/blog/feed.xml" do
     posts = PostQuery.new.all.first(10)
-    xml render_feed(posts)
-  end
-
-  private def render_feed(posts)
-    items = posts.map do |post|
-      url = Blog::Show.with(post.slug).url
-      RSS::Item.new(
-        guid: RSS::Guid.new(value: url, is_permalink: true),
-        title: post.title,
-        link: url,
-        description: post.html_content,
-        pub_date: post.published_at,
-      )
-    end
-    last_build_date = posts.map(&.published_at).max
-
-    feed = RSS::Channel.new(
-      title: "Lucky Blog",
-      description: "Lucky is a web framework written in Crystal",
-      link: Blog::Index.url,
-      feed_url: Blog::Feed.url,
-      items: items,
-      last_build_date: last_build_date,
-    )
-
-    feed.to_xml
+    xml RSS::Channel.new(posts).to_xml
   end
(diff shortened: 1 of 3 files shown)


jwoertink:
This is pretty sweet! From the first pass, it's all looks good to me. I want to take another look here in a bit. Thanks for tackling this.

src/actions/blog/feed.cr
```diff
@@ -0,0 +1,31 @@
+class Blog::Feed < BrowserAction
+  get "/blog/feed.xml" do
+    posts = PostQuery.new.all[0...10]
```
paulcsmith: There is a cool way to use `first` with an arg that I think would be great to use here:

```suggestion
    posts = PostQuery.new.all.first(10)
```

src/actions/blog/feed.cr
```diff
@@ -0,0 +1,31 @@
...
+    items = posts.map do |post|
+      url = Blog::Show.with(post.slug).url
+      RSS::Item.new(
+        guid: RSS::Guid.new(value: url, is_permalink: true),
+        title: post.title,
+        link: url,
+        description: post.html_content,
+        pub_date: post.published_at,
+      )
+    end
```
paulcsmith: I'm thinking that some of this could be moved to `RSS:Channel` and that class could be responsible for rendering times, rather than rendering them and passing them in. I'm imagining something like this:

```crystal
RSS::Channel.new(posts: posts).to_xml
```

Then `RSS::Channel` would build the `RSS::Item` with the passed in posts, private method could be added for `last_build_date`, `url`, etc. I also think the title and description of the channel could be hardcoded and not passed in since this is only used for the Lucky Blog. We could make those arguments later if we decide to add more RSS feeds.

Does this make sense? What do you think?

src/components/shared/layout_head.cr
```diff
@@ -31,4 +32,8 @@ class Shared::LayoutHead < BaseComponent
     meta property: "og:locale", content: "en_US"
     meta property: "og:image", content: Lucky::RouteHelper.settings.base_uri + asset("images/lucky_og_screenshot.png")
   end
+
+  private def rss_link
```
paulcsmith, line 36: I love nice private methods like this! 👍

paulcsmith (approved):
This is *awesome* thank you so much. Just had a refactoring suggestion, but overall looks great

wezm:
I extracted the RSS generation from an app I'm working on that generates many feeds, which is why it's generic. Happy to make those changes.

src/actions/blog/feed.cr
wezm: Ah I tried `.take(10)` but that didn't work so resorted to the range.

paulcsmith: Oh bummer! Thought it did but this is a-ok as-is

wezm:
I've made the requested changes. I made them as separate commits but happy to rebase and squash into previous commits after review.

jwoertink (approved):
Looks great to me! Thanks for adding this is.

jwoertink:
Deployed! https://luckyframework.org/blog/feed.xml

wezm:
Subscribed :tada: 

Screenshot from 2019-10-09 07-51-50

PR #181, opened 2019-10-11
Typo fix and release flag

russ:
Fixing a typo in the guide and adding a `--release` flag to the luck_cli build reference.

### DIFF
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -77,7 +77,7 @@ class Guides::GettingStarted::Installing < GuideAction
     * cd into the newly cloned directory
     * Check out latest [released version](https://github.com/luckyframework/lucky_cli/releases) `git checkout #{LuckyCliVersion.current_tag}`
     * Run `shards install`
-    * Run `crystal build src/lucky.cr`
+    * Run `crystal build --release src/lucky.cr`
     * Move the generated `lucky` binary to your path. Most of the time you can move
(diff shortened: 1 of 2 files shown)


src/actions/guides/getting-started/installing.cr
```diff
@@ -77,7 +77,7 @@ class Guides::GettingStarted::Installing < GuideAction
     * cd into the newly cloned directory
     * Check out latest [released version](https://github.com/luckyframework/lucky_cli/releases) `git checkout #{LuckyCliVersion.current_tag}`
     * Run `shards install`
-    * Run `crystal build src/lucky.cr`
+    * Run `crystal build --release src/lucky.cr`
```
jwoertink: @paulcsmith I thought this should be here, but what's your thoughts on this? Not sure if there's a reason to have this in a non-release build.

paulcsmith: The main reason was that it takes forever to build in release mode and the CLI does so little that the performance optimizations aren't noticeable. So I think we can keep it as it was

```suggestion
    * Run `crystal build src/lucky.cr`
```

jwoertink (approved):

src/actions/guides/getting-started/installing.cr
jwoertink: OMG. github... Apparently I can't commit that suggestion, and it keeps throwing an error when I try to unapprove 😝

jwoertink (approved):

PR #190, opened 2019-10-24
Rename form to operation

vlazar:
Ref https://github.com/luckyframework/lucky/issues/946

### DIFF
--- a/src/actions/guides/database/validating-saving.cr
+++ b/src/actions/guides/database/validating-saving.cr
@@ -68,7 +68,7 @@ class Guides::Database::ValidatingSaving < GuideAction
       if user # the user was saved
         html Users::ShowPage, user: user
       else
-        html Users::NewPage, operation: operation
+        html Users::NewPage, save_user: operation
       end
     end
     ```
@@ -119,17 +119,17 @@ class Guides::Database::ValidatingSaving < GuideAction
     ```crystal
     # src/pages/users/new_page.cr
     class Users::NewPage < MainLayout
-      needs operation : SaveUser
+      needs save_user : SaveUser
 
       def content
-        render_form(@operation)
+        render_form(@save_user)
       end
 
-      private def render_form(op)
+      private def render_form(operation)
         form_for Users::Create do
-          label_for op.name
-          text_input op.name
-          errors_for op.name
+          label_for operation.name
+          text_input operation.name
+          errors_for operation.name
 
           submit "Save User"
         end
@@ -150,7 +150,7 @@ class Guides::Database::ValidatingSaving < GuideAction
             redirect to: Home::Index
           else
             # re-render the NewPage so the user can correct their mistakes
-            html NewPage, user_operation: operation
+            html NewPage, save_user: operation
           end
         end
       end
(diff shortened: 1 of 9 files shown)


src/actions/guides/database/validating-saving.cr
```diff
@@ -145,12 +145,12 @@ class Guides::Database::ValidatingSaving < GuideAction
...
       route do
         # params will have the form params sent from the HTML form
-        SaveUser.create(params) do |form, user|
+        SaveUser.create(params) do |operation, user|
           if user # if the user was saved
             redirect to: Home::Index
           else
             # re-render the NewPage so the user can correct their mistakes
-            html NewPage, user_form: form
+            html NewPage, user_operation: operation
```
jwoertink: On line 122, this `NewPage` actually `needs operation`. So this should just be `operation`. However, maybe it should say `save_user: operation`, and then line 122 changes to `needs save_user : SaveUser`?

jwoertink (dismissed):
Looks great! I only noticed one thing.

src/actions/guides/database/validating-saving.cr
vlazar: Thanks for catching `needs operation` part!

And I've also had similar thoughts when noticed `save_operation: operation` in other place, while all other examples just using `operation: operation`.

Maybe for some operations it would make sense to use  a name different from just `operation` to clarify things.
Maybe always doing `operation: operation` is fine, at leas in all provided examples.

I think we need to capture a list of used names and vote which option looks best in each case.

jwoertink: I personally don't like `operation: operation` because it almost feels like you're locked in to that. I think changing it up from example to example helps to make it clear that this keyword is determined by you (the developer), and not the framework.

jwoertink: Maybe @paulcsmith or @edwardloveall have some insight here ^

edwardloveall: I think almost always it's better to have a more specific name instead of `operation`. I like the pattern we've used before of `verb_noun` like `save_user`. It would be great to find examples that made sense in context to further hint to readers what an operation does.

paulcsmith: I agree. In my own projects I always use the operation name. The generators use `operation: operation` because it was easier but in general I use `needs verb_noun`. I'm 100% ok using `operation` in the block arg though, or even `op` since it is used so much.

I think for examples though using `operation` is better for clarity. So I like this change 👍

src/actions/guides/frontend/rendering_html.cr
```diff
@@ -705,11 +705,11 @@ class Guides::Frontend::RenderingHtml < GuideAction
     ```crystal
     # In src/posts/new_page.cr
     class Posts::NewPage < MainLayout
-      needs form : SavePost
+      needs operation : SavePost
```
paulcsmith: Maybe this can be `save_post` to match what we discussed above

src/actions/guides/frontend/rendering_html.cr
```diff
@@ -705,11 +705,11 @@ class Guides::Frontend::RenderingHtml < GuideAction
...
 
       def content
         h1 "New Blog Post"
-        render_post_form(@form)
+        render_post_form(@operation)
       end
 
       def render_post_form(f)
```
paulcsmith: The `f ` should be renamed to `operation`

paulcsmith (changes requested):
Very nice. So glad these are getting fixed 👍

src/actions/guides/database/validating-saving.cr
vlazar: Updated to suggested more clear names.

paulcsmith (approved):
Looks great! Thank you

PR #195, opened 2019-11-26
Added extra link for updating records

### LINKED ISSUE #194: Cleaning up docs on updating records
I just got bit by this. When looking up how to update records, you'll go to [this guide](https://luckyframework.org/guides/database/validating-saving#updating-records) first. The first part is we mention `params`, but don't really say what `params` looks like (i.e. `Avram::Params.new` vs `Lucky::Params.new`). When writing a spec, I assumed I had to pass in an avram params object as my second argument. In doing that, the record saves, but doesn't actually update. No errors, just no updates. 

Now, we actually do document this [here](https://luckyframework.org/guides/database/validating-saving#saving-without-a-params-object); however, it's a LOT further down the page. I didn't come across it until after a half hour of trying to setup postgres logs locally 😅 

We can easily just add a link below the updating that says "If you need to save without a params object, go here" (or whatever). Also might be helpful to explain what that `params` object is.

jwoertink:
Fixes #194

Just a small note right after the updating record docs to point to more information further down the page.

### DIFF
--- a/src/actions/guides/database/validating-saving.cr
+++ b/src/actions/guides/database/validating-saving.cr
@@ -105,7 +105,8 @@ class Guides::Database::ValidatingSaving < GuideAction
     ```
 
     > `params` is defined in your actions for you. You can also
-    > [save without a params object](#saving-without-a-params-object), for example, in your specs.
+    > [save without a params object](#saving-without-a-params-object), for example, in your specs, or in
+    > a seeds file.
 
     ## Using with JSON endpoints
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/validating-saving.cr
```diff
@@ -104,6 +104,9 @@ class Guides::Database::ValidatingSaving < GuideAction
     updated_user = SaveUser.update!(user, params)
     ```
 
+    > `params` is defined in your actions for you. You can also
+    > [save without a params object](#saving-without-a-params-object), for example, in your specs.
```
paulcsmith: And maybe include it’s helpful in seeds?

paulcsmith (approved):
Love it! Just one suggestion but feel free to merge without if you think it’s too much

PR #197, opened 2019-12-04
test gzip setup

bdtomlin:
Testing gzip for https://github.com/luckyframework/lucky/issues/154
Gzip is handled two different ways which can be enabled or disabled in `config/gzip.cr`:

For static assets:
* Webpack uses `compression-webpack-plugin` to pre-compress files
*  `Lucky::StaticCompressionHandler` attempts to serve compressed file then calls next middleware if it's not available or it's not a supported file type.

For server generated responses:
* `Lucky::DynamicCompressionHandler` module can be mixed in to middleware that needs to serve compressed responses. See `src/gzip/dynamic_compression_handler.cr` for temporarily monkey patched versions of `Lucky::RouteHandler` and `Lucky:: RouteNotFoundHandler`
* The `Lucky::DynamicCompressionHandler#gzip` method wraps the `context.response.output` IO with a Gzip::Writer.

I've only added gzip in this example. Adding brotli to the static asset compression is trivial, but I think this should be left off for simplicity, especially since I'm not sure what would be involved in making that work for server generated responses.


src/gzip/dynamic_compression_handler.cr
```diff
@@ -0,0 +1,57 @@
+{% if !flag?(:without_zlib) %}
+  require "gzip"
+{% end %}
+
+module Lucky::DynamicCompressionHandler
+  COMPRESS_ONLY_THESE_EXTENSIONS = ["", ".html", ".htm", ".txt", ".json", ".xml"]
```
jwoertink, line 6: Why is there a blank one at the beginning? Is that like for paths with no extension (default)?

bdtomlin: Yep.

config/gzip.cr
```diff
@@ -0,0 +1,10 @@
+Lucky::DynamicCompressionHandler.configure do |settings|
+  settings.compress_server_generated_content = true
+end
+
+# This setting turns off the serving of pre-compressed assets, but they will still
+# be generated by webpack. To disable this remove the compression-webpack-plugin
+# in webpack.mix.js
+Lucky::StaticCompressionHandler.configure do |settings|
```
paulcsmith, line 8: I like this name 👍

config/gzip.cr
```diff
@@ -0,0 +1,10 @@
...
+  settings.serve_precompressed_assets = true
```
paulcsmith, line 9: I think this  could be `enabled = true`

config/gzip.cr
```diff
@@ -0,0 +1,10 @@
+Lucky::DynamicCompressionHandler.configure do |settings|
+  settings.compress_server_generated_content = true
```
paulcsmith, line 2: This could be `enabled`

src/app_server.cr
```diff
@@ -9,6 +11,7 @@ class AppServer < Lucky::BaseAppServer
       Lucky::ErrorHandler.new(action: Errors::Show),
       Lucky::RouteHandler.new,
       CacheControlHandler.new,
+      Lucky::StaticCompressionHandler.new("./public", file_ext: "gz", content_encoding: "gzip"),
```
paulcsmith, line 14: I'm thinking all 3 of these could be Habitat settings instead

src/gzip/dynamic_compression_handler.cr
paulcsmith, line 6: This could be a habitat setting as well, that way people can configure it if desired

src/gzip/static_compression_handler.cr
```diff
@@ -0,0 +1,79 @@
+class Lucky::StaticCompressionHandler
+  Habitat.create do
+    setting serve_precompressed_assets : Bool = false
+  end
+
+  include HTTP::Handler
+
+  GZIP_FILE_EXTENSIONS = %w(.htm .html .txt .css .js .svg .json .xml .otf .ttf .woff .woff2)
```
paulcsmith, line 8: Nice! I was wondering about this. Could you make this a Habitat setting as well?

webpack.mix.js
```diff
@@ -44,7 +45,13 @@ mix
   // Reduce noise in Webpack output
   .webpackConfig({
     stats: "errors-only",
-    plugins: [webpackNotifier]
+    plugins: [
+      webpackNotifier,
+      new CompressionPlugin({
+        compressionOptions: { level: 9 },
+        test: /\.js$|\.css$|\.html$|\.svg$/
```
paulcsmith, line 52: 🎉 Nice and simple. I love it! I wonder if we could only enable this in production though? Maybe not worth worrying about now, but just a thought

paulcsmith (commented):
Tried it out. Works fantastically well as far as I can tell! I added feedback here but no need to address it in this PR since I'm guessing you'll make a PR to Lucky and Lucky CLI to add this?

I'm really excited about this. Massive network savings in the JS and CSS for the Lucky website ❤️

src/gzip/dynamic_compression_handler.cr
paulcsmith: Actually I just realized that  maybe rather than compressing using an extension it might be more reliable to use the `Content-Type` header. Thoughts?

I think this should work since by default Lucky adds the correct content-type for most common responses instead of setting an extension: https://github.com/luckyframework/lucky/blob/ed8d36658b7f83622dad09e81582575bc164c564/src/lucky/renderable.cr#L85

config/gzip.cr
bdtomlin: I agree, that's much better.

config/gzip.cr
bdtomlin: I agree, that's much better.

paulcsmith:
Seriously so stoked about this. I love how this turned out!!

src/app_server.cr
bdtomlin: I don't think these should be habitat settings. I think if you were to make changes you would probably add brotli, but in that case you still need the gzip fall through.

```crystal
Lucky::StaticCompressionHandler.new("./public", file_ext: "br", content_encoding: "br"),
Lucky::StaticCompressionHandler.new("./public", file_ext: "gz", content_encoding: "gzip"),
```

src/gzip/dynamic_compression_handler.cr
bdtomlin: I have to think about that. I'm not sure the size of the footgun. Mostly I'm thinking about binary that shouldn't be compressed. I'm wondering if it should be the framework's responsibility to know what is reasonable to compress? I'm a bit on the fence.

src/gzip/static_compression_handler.cr
bdtomlin: I have to think about that. I'm not sure the size of the footgun. Mostly I'm thinking about binary that shouldn't be compressed. I'm wondering if it should be the framework's responsibility to know what is reasonable to compress? I'm a bit on the fence.

paulcsmith: @bdtomlin Oh sorry I think maybe I wasn't clear. I meant something like this so we still give a nice default:

```crystal
setting gzip_file_extensions = %w(.htm .html .txt .css .js .svg .json .xml .otf .ttf .woff .woff2)
```

But users can change it or add to it if they want:

```crystal
StaticCompressionHandler.configure do |config|
  config.gzip_file_extensions << ".somethingelse"
end
```

So I think with this covers the default case but if users are having issues or need to add another type, they can

webpack.mix.js
bdtomlin: So in my testing there hasn't been any noticeable performance hit, I'll definitely defer to whatever you want on this since I don't have a strong opinion.

src/app_server.cr
paulcsmith: Oh interesting. So you could have multiple StaticCompressionHandlers that handle different file types?

paulcsmith: If so then yeah I totally agree with you and that's very cool!

webpack.mix.js
paulcsmith: Let's keep it as-is. Definitely not worth slowing this down. People can tweak it if it starts to become an issue 👍

src/gzip/static_compression_handler.cr
bdtomlin: Yeah, I'm persuaded. Will do.

src/app_server.cr
bdtomlin: Exactly.

Just add
```js
    plugins: [
      webpackNotifier,
      new CompressionPlugin({
        compressionOptions: { level: 9 },
        test: /\.js$|\.css$|\.html$|\.svg$/
      }),
      new CompressionPlugin({
        compressionOptions: { level: 11 },
        filename: '[path].br[query]',
        algorithm: 'brotliCompress',
        test: /\.js$|\.css$|\.html$|\.svg$/
      })
    ],
```
to your webpack.mix.js and you're rolling.

src/gzip/dynamic_compression_handler.cr
bdtomlin: will do

src/app_server.cr
paulcsmith: That is sooo cool. I love it

src/gzip/dynamic_compression_handler.cr
bdtomlin: I don't think Content-Type has been set yet at this point. Basically we are wrapping the IO with the Gzip::Writer before the render happens. Let me do a little testing to make sure I'm right about that.

bdtomlin: Yeah, I was right about that. We only have headers available from the request at this point.

paulcsmith: Ah I see. Ok I have a proposal if you are up for it. We could do 2 PRs. Once for static file handler and one for the dynamic one. My hunch is the static one will be simpler and we can get that in pretty quick. The dynamic one may be a bit trickier to really nail down but if it is a separate PR it won't block the static one.

Thoughts?

paulcsmith: Another idea is that we set the GZIP writer when printing the text response instead of doing it in a handler. https://github.com/luckyframework/lucky/blob/master/src/lucky/text_response.cr

bdtomlin: I agree with 2 PRs. I had thought the same thing, but I just did the one PR here to give an idea of the combined strategy.

I'll look into setting the writer in that location before doing the PR for the dynamic handler.

paulcsmith: Oh perfect! Sounds like you've already got this done. Stoked to see the PRs. Thanks for thoroughly researching all the options. I like this a lot more than where we were previously going with doing gzip at runtime 👍. Also cool that people can use other compression like Brotli if they want

paulcsmith:
Closing now that this is in Lucky and Lucky CLI master!

PR #207, opened 2019-12-28
Update Migrations page, clarify  default values and fill_existing_with

btihen:
Migration Guide update

Clarifies how default and fill_existing_with works and fixes an incompatibility in the example code.

### DIFF
--- a/src/actions/guides/database/migrations.cr
+++ b/src/actions/guides/database/migrations.cr
@@ -258,7 +258,7 @@ class Guides::Database::Migrations < GuideAction
 
     When using the `add` method inside an `alter` block, there's an additional option `fill_existing_with`.  This is useful when a default value is **NOT** desired, but the existing data should be backfilled anyway.
 
-    If your column is required, you will need to set a **default** value on all records otherwise you'll have errors.  **NOTE:** `default` also back-fills existing data just like `fill_existing_with` and may not be used in combination with the `default` option.
+    If your column is required, you will need to set a **default** value on all records or use `fill_existing_with` otherwise you'll have errors.  **NOTE:** You must use either `default` or  `fill_existing_with`. They can't be used together since they both solve similar problems.
 
     ```crystal
     alter table_for(User) do
--- a/src/actions/guides/database/migrations.cr
+++ b/src/actions/guides/database/migrations.cr
@@ -254,15 +254,15 @@ class Guides::Database::Migrations < GuideAction
 
     > You can also use a symbol for a table name. For example `create :users`.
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/migrations.cr
```diff
@@ -254,15 +254,15 @@ class Guides::Database::Migrations < GuideAction
 
     > You can also use a symbol for a table name. For example `create :users`.
 
-    ## Using fill_existing_with
+    ## Using fill_existing_with and default values
 
-    When using the `add` method inside an `alter` block, there's an additional option `fill_existing_with`.
+    When using the `add` method inside an `alter` block, there's an additional option `fill_existing_with`.  This is useful when a default value is **NOT** needed, but the existing data should be backfilled anyway.
```
paulcsmith:
```suggestion
    When using the `add` method inside an `alter` block, there's an additional option `fill_existing_with`.  This is useful when a default value is **NOT** desired, but the existing data should be backfilled anyway.
```

src/actions/guides/database/migrations.cr
```diff
@@ -254,15 +254,15 @@ class Guides::Database::Migrations < GuideAction
...
 
-    If your column is required, you will need to set a default value on all records otherwise you'll have errors.
+    If your column is required, you will need to set a **default** value on all records otherwise you'll have errors.  **NOTE:** `default` also back-fills existing data just like `fill_existing_with` and may not be used in combination with the `default` option.
```
paulcsmith:
```suggestion
    If your column is required, you will need to set a **default** value on all records or use `fill_existing_with` otherwise you'll have errors.  **NOTE:** You must use either `default` or  `fill_existing_with`. They can't be used together since they both solve similar problems.
```

The default doesn't really "backfill" data. It's just that you don't need `fill_existing_with` if you have a default since they serve similar pruposes. If you have a default then you don't need fill_existing_with. Does that make sense?

src/actions/guides/database/migrations.cr
paulcsmith: Just changed to `desired` since it is more about what you want to do. You could do either default or fill_existing_with depending on what you want. `fill_existing_with` is usually used when you don't want a default. You want ever new record to specify a value for the column. `default` is better when you have a sensible default (like for an `admin` column with a default of `false`)

paulcsmith (commented):
Thank you! I'm hoping I clarified this a bit. Maybe you have ideas for further clarification in the guide based on my comments? Otherwise I think this looks good!

btihen:
Thanks for clarifying suggestions.

Issue #212, opened 2020-02-18
Document new select_sum! and the return types for select_sum


paulcsmith:
https://github.com/luckyframework/avram/pull/304

`select_sum!` returns `0` if there are no records to sum. `select_sum` returns `nil`. Integer column types return an Int64. Floats return a Float.

Guide link: https://github.com/luckyframework/website/blob/master/src/actions/guides/database/querying-deleting.cr#L358

bdtomlin:
I'll try to get to this tonight or tomorrow while it's still fresh in my head.

paulcsmith:
That would be wonderful! Thanks @bdtomlin

jwoertink:
Looks like this is already in and deployed. Thanks @bdtomlin!

PR #213, opened 2020-02-21
Add documentation for aggregate queries

bdtomlin:
#212

### DIFF
--- a/src/actions/guides/database/querying-deleting.cr
+++ b/src/actions/guides/database/querying-deleting.cr
@@ -400,7 +400,7 @@ class Guides::Database::QueryingDeleting < GuideAction
     ```
 
     `select_min` and `select_max` will return a union type of the column and `Nil`.
-    > For example, if the column type is an `Int32` the return type will be `Int32 | Nil`.
+    For example, if the column type is an `Int32` the return type will be `Int32 | Nil`.
 
 
     ## Associations and Joins
--- a/src/actions/guides/database/querying-deleting.cr
+++ b/src/actions/guides/database/querying-deleting.cr
@@ -360,14 +360,30 @@ class Guides::Database::QueryingDeleting < GuideAction
     `SELECT AVG(users.age) FROM users`
 
     ```crystal
+    # This will return a Float64 | Nil.
+    # The value will be nil if there are no records.
     UserQuery.new.age.select_average
+
+    # This will return a Float64.
+    # The value will be 0 if there are no records.
+    UserQuery.new.age.select_average!
     ```
 
     `SELECT SUM(users.age) FROM users`
 
     ```crystal
+    # Returns Nil if there are no records
     UserQuery.new.age.select_sum
+
+    # Returns 0 if there are no records
+    UserQuery.new.age.select_sum!
     ```
+
+    > The return type for `select_sum` will be `Int64 | Nil` if the column is an `Int`.
+    > If the column is a `Float` it will be `Float64 | Nil`.
+    >
+    > The return type for `select_sum!` will be `Int64` if the column is an `Int`.
+    > If the column is a `Float` it will be `Float64`.
 
     ### Min / Max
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/querying-deleting.cr
```diff
@@ -360,15 +360,31 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+
+    # This will return a Float64.
+    # The value will be 0 if there are no records.
+    UserQuery.new.age.select_average!
     ```
 
     `SELECT SUM(users.age) FROM users`
 
     ```crystal
+    # Returns Nil if there are no records
```
paulcsmith: Maybe this could be split into two lines and with a bit more detail:

```
# Returns an Int64 for integer columns, or a Float64 for float columns
# Returns 'nil' if there are no records
```

src/actions/guides/database/querying-deleting.cr
```diff
@@ -360,15 +360,31 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+
+    # Returns 0 if there are no records
+    UserQuery.new.age.select_sum!
     ```
 
+    > The return type for `select_sum` will be `Int64 | Nil` if the column is an `Int`.
+    > If the column is a `Float` it will be `Float64 | Nil`.
+    >
+    > The return type for `select_sum!` will be `Int64` if the column is an `Int`.
+    > If the column is a `Float` it will be `Float64`.
```
paulcsmith: I think this could go before the code sample and probably doesn't need `>` since I think it is pretty essential knowledge. Thoughts?

src/actions/guides/database/querying-deleting.cr
```diff
@@ -383,6 +399,10 @@ class Guides::Database::QueryingDeleting < GuideAction
     UserQuery.new.age.select_max
     ```
 
+    > `select_min` and `select_max` will return a union type of the column and `Nil`.
```
paulcsmith:
```suggestion
    `select_min` and `select_max` will return a union type of the column and `Nil`.
```

src/actions/guides/database/querying-deleting.cr
```diff
@@ -383,6 +399,10 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+    > For example, if the column type is an `Int32` the return type will be `Int32 | Nil`.
```
paulcsmith:
```suggestion
    For example, if the column type is an `Int32` the return type will be `Int32 | Nil`.
```

paulcsmith (commented):
This is fantastic. Very well explained. Thank you! Just a few small suggestions

src/actions/guides/database/querying-deleting.cr
bdtomlin: TBH if found the use of `>` a little confusing. I figured the block quotes were just being used to call attention because their styles really stand out. I think it's fine either way. I don't mind moving it to the top.

paulcsmith: Yeah I think I overused block quotes at first 😆. The rough guideline is if there is information that *might* be helpful or to call out best practices that are tangentially related.

bdtomlin:
Makes sense.

src/actions/guides/database/querying-deleting.cr
bdtomlin: Actually, I think this should just be deleted. Your other suggestion for a better comment in the code example covers this anyways.

paulcsmith:
I don't know why CI is failing, but I tried this locally and it worked so I'm going to go ahead and merge and deploy. Thank you! It looks great

bdtomlin:
@paulcsmith CI is failing because there is no .env file. An empty .env in ci will fix the issue.

paulcsmith:
It's just strange because it was working before.I wonder if dotenv changed something? I'll have to look into why it is doing that. I'd rather have it so that dotenv continues if no .env is found

bdtomlin:
@paulcsmith this currently happens in any environment. I ran into it when setting up my production server.

paulcsmith:
Oh dang :S Yeah we need to change that. If you don't have a .env it should just carry on

paulcsmith:
Fixed in https://github.com/luckyframework/website/pull/214 and https://github.com/luckyframework/lucky_cli/pull/451

PR #219, opened 2020-02-29
Adding docs for nilable_eq. 

### LINKED ISSUE #182: Missing docs on nilable_eq
Or I could be blind... but I couldn't find them. 

https://github.com/luckyframework/avram/blob/master/src/avram/criteria.cr#L63

We should add to https://luckyframework.org/guides/database/querying-deleting#where-queries somewhere.

jwoertink:
Fixes #182

### DIFF
--- a/src/actions/guides/database/querying-deleting.cr
+++ b/src/actions/guides/database/querying-deleting.cr
@@ -171,6 +171,15 @@ class Guides::Database::QueryingDeleting < GuideAction
     UserQuery.new.age(54)
     ```
 
+    In most cases, your value will come from a dynamic source like `params`.
+    If the value you pass in to your query method is nilable, Avram will raise an exception.
+
+    For these cases, you can use the `nilable_eq` method.
+
+    ```crystal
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/querying-deleting.cr
```diff
@@ -171,6 +171,15 @@ class Guides::Database::QueryingDeleting < GuideAction
     UserQuery.new.age(54)
     ```
 
+    In most cases, your value will come from a dynamic source like `params`.
```
paulcsmith: I’d maybe say in “some” cases since it really depends on the app. And maybe switch the sentences so it says “if the value...” since it may happen with non params and the most important part is the nil

paulcsmith (commented):
Just one comment. Looks good to merge when you’re ready

src/actions/guides/database/querying-deleting.cr
jwoertink: I originally wrote "some", and then changed it because I figured maybe it was edge case 😅  I've re-worded it now

paulcsmith (approved):
Looks good!

PR #225, opened 2020-03-04
Make common things a bit more visible in search

paulcsmith:
By using words people use. Not the indexer does not index code, so using regular words is important


jwoertink (approved):
Ah yeah. That makes sense from a searching standpoint.

PR #227, opened 2020-03-05
Work on a new HTML forms page.

### LINKED ISSUE #183: Update mention of errors_for
On [this page](https://luckyframework.org/guides/database/validating-saving#using-with-html-forms) there's a mention of `errors_for`

```crystal
label_for op.name
text_input op.name
errors_for op.name
```

which actually should use [this component](https://github.com/luckyframework/lucky_cli/blob/master/src/browser_app_skeleton/src/components/shared/field.cr)  or at least just update to use [FieldErrors](https://github.com/luckyframework/lucky_cli/blob/master/src/browser_app_skeleton/src/components/shared/field_errors.cr)

```crystal
mount Shared::Field.new(op.name)
# or

label_for op.name
text_input op.name
mount Shared::FieldErrors.new(op.name)
```

### LINKED ISSUE #193: Document all form inputs
There's lots of input methods like `datetime_input`, and `telephone_input` that don't seem to be documented on the site. We should make sure there's a whole section dedicated to form input html methods.

jwoertink:
Fixes #193
Fixes #183 

There's a lot going on here, so I don't want to merge just yet, but please look over it and add suggestions. 

The current documentation for building HTML forms is located in the "Validating and Saving" section of "Database". It kind of makes sense because you use forms to save data, and we tie this stuff in to Operations. However, lots of people new to Lucky won't think "How do I do a select form? oh, let me look under database...". Since this is HTML related, I wanted to dedicate an entire page on the front end section to showing all of the methods we have, and how to use them. 

This will be the initial version of this page, but eventually I see this including more complex examples, and talking about doing associations and such.

I've also removed *some* of the html mentions from the validating page, though, I think because there's cross-over between them, it's ok to show some parts. It's like the HTML forms page shows the html side, and then links to the operations, where the validations page shows from the operations side and links to the HTML page. 

Last note was we had some mentions of things like `errors_for` which doesn't exist anymore

### DIFF
--- a/src/actions/guides/frontend/html_forms.cr
+++ b/src/actions/guides/frontend/html_forms.cr
@@ -319,6 +319,9 @@ class Guides::Frontend::HtmlForms < GuideAction
 
     ## Select fields
 
+    For select fields, you'll use a combination of `select_tag` and the `options_for_select` methods.
+    The selected value will be determined by the current value of the attribute (i.e. `op.car_make`)
+
     ```html
     <select name="param_key:car_make" class="custom-select">
       <option value="1">Honda</option>
@@ -329,14 +332,25 @@ class Guides::Frontend::HtmlForms < GuideAction
     ```
 
     ```crystal
-    select(op.car_make, class: "custom-select") do
+    select_tag(op.car_make, class: "custom-select") do
       options_for_select(op.car_make, [{"Honda", 1}, {"Toyota", 2}, {"Ford", 3}, {"Volkswagen", 4}])
     end
     ```
 
-    ## Checboxes / Radios
+    The second argument to `options_for_select` takes an `Array(Tuple(String, String | Int32 | Int64))`. If your data is coming from a query, you can easily map that data in to this format.
 
-    ### Chechbox
+    ```crystal
+    # Use with `options_for_select(op.car_make, options_for_cars)`
+    private def options_for_cars
+      CarQuery.all.map do |car|
+        {car.make, car.id}
+      end
+    end
+    ```
+
+    ## Checkboxes / Radios
+
+    ### Checkbox
 
     The `checkbox` method will auto generate a secondary hidden input that will hold the
     unchecked value of your attribute. This allows params to still contain a value even if a
--- a/src/actions/guides/frontend/html_forms.cr
+++ b/src/actions/guides/frontend/html_forms.cr
@@ -7,7 +7,7 @@ class Guides::Frontend::HtmlForms < GuideAction
 
   def markdown : String
     <<-MD
-    ### Overview
+    ## Overview
 
     You can generate form tags like `<form>`, `<input>`, etc... using
     [methods like `form` and `input`](#{Guides::Frontend::RenderingHtml.path}),
--- a/src/actions/guides/frontend/html_forms.cr
+++ b/src/actions/guides/frontend/html_forms.cr
@@ -47,7 +47,7 @@ class Guides::Frontend::HtmlForms < GuideAction
     The Action class already has a route defined that defines which HTTP method to use (GET, PUT, etc.). When passing
     the action class in, Lucky will use the defined HTTP method automatically.
 
-    You can also pass in a route path if you need to pass in additional URL params.
+    You can also pass in a route path using `with` if the action requires params to be passed in.
 
     ```crystal
     form_for(Posts::Create.with(author_id: @current_user.id)) do
--- a/src/actions/guides/frontend/html_forms.cr
+++ b/src/actions/guides/frontend/html_forms.cr
@@ -44,7 +44,7 @@ class Guides::Frontend::HtmlForms < GuideAction
     end
     ```
 
-    The Action class already has a route defined that sets the HTTP METHOD. When passing
+    The Action class already has a route defined that defines which HTTP method to use (GET, PUT, etc.). When passing
     that class in, Lucky will know which method to set for you.
 
     You can also pass in a route path if you need to pass in additional URL params.
--- a/src/actions/guides/frontend/html_forms.cr
+++ b/src/actions/guides/frontend/html_forms.cr
@@ -66,7 +66,7 @@ class Guides::Frontend::HtmlForms < GuideAction
 
     ## Input fields
 
-    All of the input helper methods take an `Avram::Attribute`. These are created
+    All of the input helper methods take an `Avram::PermittedAttribute`. These are created
     from setting an `attribute` or `permit_columns` in your `Avram::Operation`.
     Then calling that attribute on the Operation.
     See the [Operations Guide](#{Guides::Database::ValidatingSaving.path}) for more info.
--- a/src/actions/guides/frontend/html_forms.cr
+++ b/src/actions/guides/frontend/html_forms.cr
@@ -45,7 +45,7 @@ class Guides::Frontend::HtmlForms < GuideAction
     ```
 
     The Action class already has a route defined that defines which HTTP method to use (GET, PUT, etc.). When passing
-    that class in, Lucky will know which method to set for you.
+    the action class in, Lucky will use the defined HTTP method automatically.
 
     You can also pass in a route path if you need to pass in additional URL params.
 
--- a/src/actions/guides/frontend/html_forms.cr
+++ b/src/actions/guides/frontend/html_forms.cr
@@ -405,15 +405,21 @@ class Guides::Frontend::HtmlForms < GuideAction
 
     ## Labels
 
+    The text will be derived from the name of the attribute.
+
     ```crystal
-    label_for(op.first_name)
+    label_for(op.first_name, class: "custom-label")
+    ```
 
-    # or with custom text
+    If you need custom text, you can pass it in as the second argument.
 
-    label_for(op.first_name, "Enter your First name:")
+    ```crystal
+    label_for(op.first_name, "Enter your First name:", class: "custom-label")
+    ```
 
-    # or using a block
+    or use a block for extra customization
 
+    ```crystal
     label_for(op.first_name) do
       strong("First Name: ")
     end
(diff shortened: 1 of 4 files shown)


paulcsmith (commented):
Still need to do a deeper dive but overall I love the direction! Linking between the two is a great idea and the examples are good. I love this improvement!

jwoertink:
Ok, I've done a deep dive on this, and I'm happy with the direction. Let me know if you see any typos, or wording clarity on anything.

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+
+    All HTML tags have an associated method based on the name of the tag, with the exception of a few.
+    (e.g. `<form></form>` - `form()`, `<button></button>` - `button()`, etc...)
+
+    These are the only HTML tags with an alias name method.
+
+    * `<select></select>` - `select_tag`
+    * `<p></p>` - `para`
+
+    > These are overridden as to not clash with built-in Crystal methods.
```
paulcsmith: This seems like a general overview which feels odd to have duplicated from the rendering html guide. I'm thinking the first part of this paragraph could be something like:

> You can generate form tags like `<form>`, `<input>`, etc. using [methods like `form` and `input`]#{link to rendering html}}, but the recommended way is to use Lucky's form helpers.

Then lead with the rest of what you have

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
+class Guides::Frontend::HtmlForms < GuideAction
+  guide_route "/frontend/html-forms"
+
+  def self.title
+    "HTML Forms"
+  end
+
+  def markdown : String
+    <<-MD
+    ### Working with Lucky's HTML builder
```
paulcsmith: Thoughts on "Overview" so it doesn't take as much space in the table of contents?

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+    ```
+
+    becomes
+
+    ```crystal
+    form_for(Posts::Create, class: "inline-form") do
+    end
+    ```
+
+    The Action class already has a route defined that sets the HTTP METHOD. When passing
```
paulcsmith:
```suggestion
    The Action class already has a route defined that defines which HTTP method to use (GET, PUT, etc.). When passing
```

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+    that class in, Lucky will know which method to set for you.
```
paulcsmith:
```suggestion
    the action class in, Lucky will use the defined HTTP method automatically.
```

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+
+    You can also pass in a route path if you need to pass in additional URL params.
```
paulcsmith:
```suggestion
    You can also pass in a route path using `with` if the action requires params to be passed in.

File: src/actions/guides/frontend/html_forms.cr [hunk 6]
```diff
@@ -0,0 +1,533 @@
...
+    method directly, and pass any options you'd like.
+
+    ```crystal
+    form(method: "get", class: "custom-form", action: Search::Index.path) do
+    end
+    ```
+
+    ## Input fields
+
+    All of the input helper methods take an `Avram::Attribute`. These are created
```
paulcsmith:
```suggestion
    All of the input helper methods take an `Avram::PermittedAttribute`. These are created
```

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+    from setting an `attribute` or `permit_columns` in your `Avram::Operation`.
```
paulcsmith:
```suggestion
    from declaring an `attribute` or `permit_columns` in your `Avram::Operation`.
```

Maybe link to the permit columns section? It has a permalink ready to go :D

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+    Then calling that attribute on the Operation.
```
paulcsmith: I'm not sure I understand this part since I don't think you need to do anything extra for an attribute to work in a form input

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+    See the [Operations Guide](#{Guides::Database::ValidatingSaving.path}) for more info.
+
+    > For these examples, `op` will refer to an instance of an `Avram::Operation`.
```
paulcsmith:
```suggestion
    > For these examples, `op` will refer to an instance of an `Avram::Operation` (such as `SaveUser`).
```

Thoughts on the example? Helpful or harmful? I'm fine either way but wanted to see what you think

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+           id="param_key_full_name"
+           name="param_key:full_name"
+           value=""
+           class="custom-input"
+           required />
+    ```
+
+    ```crystal
+    text_input(op.full_name, attrs: [:required], class: "custom-input")
+    ```
```
paulcsmith, line 86: Thoughts on moving the Lucky method first, then showing what it generates? I think that is a bit more intuitive. Like "write this and you get this"

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+      <option value="1">Honda</option>
+      <option value="2" selected="true">Toyota</option>
+      <option value="3">Ford</option>
+      <option value="4">Volkswagen</option>
+    </select>
+    ```
+
+    ```crystal
+    select_tag(op.car_make, class: "custom-select") do
+      options_for_select(op.car_make, [{"Honda", 1}, {"Toyota", 2}, {"Ford", 3}, {"Volkswagen", 4}])
```
paulcsmith: Thoughts on shortening this to 2 items so it doesn't overflow?

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+    ## Saving Data
+
+    HTML Forms in Lucky are based around the concept of [Operations](#{Guides::Database::ValidatingSaving.path}). We use these for securing param values from form inputs, and doing validations.
+
+    For info on interacting with databases, see the [saving
+    data with operations](#{Guides::Database::ValidatingSaving.path(anchor: Guides::Database::ValidatingSaving::ANCHOR_USING_WITH_HTML_FORMS)}) guide.
+
+    ## Displaying Errors
+
+    If your permitted column / attribute fail any sort of validation, it's common to
```
paulcsmith:
```suggestion
    If your permitted column / attribute fails any sort of validation, it's common to
```

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+            label_for(@op.body)
+            textarea(@op.body)
+            error_for(@op.body)
+          end
+
+          para do
+            label_for(@op.released_at)
+            datetime_input(@op.released_at)
+            error_for(@op.released_at)
+          end
```
paulcsmith: Could we shorten this to 2 or 3? I know later on it shows how to make it more concise with the Shared::Field component, but I think this could still be a tad bit shorter

paulcsmith (approved):
Some minor suggestions. This is such a huge improvement 👍

src/actions/guides/frontend/html_forms.cr
jwoertink: 😂 Yes, that's much better. I had a lack of words on that.

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,533 @@
...
+    label_for(op.first_name)
+
+    # or with custom text
+
+    label_for(op.first_name, "Enter your First name:")
+
+    # or using a block
+
+    label_for(op.first_name) do
+      text("First Name: ")
```
paulcsmith: I think a good example here would be using a `strong` tag

```suggestion
      strong("First Name: ")
```

src/actions/guides/frontend/html_forms.cr
jwoertink: What I was trying to say is how you call `@op.some_attr_thing`. You're just calling that attribute on the operation instance. Maybe irrelevant in this case though. I'll remove it and see how that reads.

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,512 @@
+class Guides::Frontend::HtmlForms < GuideAction
+  guide_route "/frontend/html-forms"
+
+  def self.title
+    "HTML Forms"
+  end
+
+  def markdown : String
+    <<-MD
+    ### Overview
```
paulcsmith:
```suggestion
    ## Overview
```

src/actions/guides/frontend/html_forms.cr
```diff
@@ -0,0 +1,512 @@
...
+    # or with custom text
+
+    label_for(op.first_name, "Enter your First name:")
+
+    # or using a block
+
+    label_for(op.first_name) do
+      strong("First Name: ")
+    end
+    ```
```
paulcsmith, line 426: This looks funny when rendered so I think maybe each example should be split into its own code block

Screen Shot 2020-03-05 at 3 53 44 PM

paulcsmith (approved):
2 small suggestions. LGTM!

src/actions/guides/frontend/html_forms.cr
jwoertink: That's weird. Why are the comments so faded on your screen?

Screen Shot 2020-03-05 at 1 07 52 PM

jwoertink: but yeah, I'll break them out in to separate blocks... but I wonder if that is another weird css bug?

jwoertink:
Sweet! Merging!

PR #229, opened 2020-03-06
Fixing security issue

jwoertink:
* upgraded all of the node packages.
* Moved to latest tailwind config setup. 

This should put us in the latest updates for all of these so it should be a good while before we need to update again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,10 @@
   "license": "UNLICENSED",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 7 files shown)


jwoertink:
oh that's interesting... I wonder why the deploy failed..

paulcsmith:
Looks good! Design looks identical 👍

paulcsmith (approved):

PR #248, opened 2020-03-18
Improve the installation section

### LINKED ISSUE #247: Add the open ssl section back to install guide
https://github.com/luckyframework/website/commit/4e2d88eac03023025f59f90ee631337ff1559021#r37885130

Turns out you do still need it :(

paulcsmith:
* Give each dependency its own section.
* Make code blocks clearer
* Clean up some wording
* Fixes some style issues

Also fixes #247

### DIFF
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -9,9 +9,57 @@ class Guides::GettingStarted::Installing < GuideAction
   def markdown : String
     <<-MD
     #{permalink(ANCHOR_INSTALL_REQUIRED_DEPENDENCIES)}
-    ## Install Required Dependencies
+    ## MacOS requirements
 
-    ### Crystal v0.33.0
+    ### 1. Install Homebrew
+
+    Installation instructions from the [Homebrew website](https://brew.sh/)
+
+    ```plain
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+    ```
+
+    ### 2. Install OpenSSL
+
+    ```plain
+    brew install openssl
+    ```
+
+    ### 3. Configure SSL for Crystal
+
+    You'll need to add `export PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig`
+    to your `~/.bash_profile` or `~/.zshrc` so Crystal can use SSL through the
+    openssl package.
+
+    ```shell
+    # For Bash, the default shell on MacOS
+    echo 'export PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig' >>~/.bash_profile
+
+    # For ZSH
+    echo 'export PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig' >>~/.zshrc
+    ```
+
+    > If you get an error like this: "Package libssl/libcrypto was not found in the
+      pkg-config search path" then be sure to run this step so that
+      Crystal knows where OpenSSL is located.
+
+    ## Linux requirements
+
+    ### Debian (Ubuntu should be similar)
+
+    ```plain
+    apt-get install libc6-dev libevent-dev libpcre2-dev libpng-dev libssl1.0-dev libyaml-dev zlib1g-dev
+    ```
+
+    ### Fedora (28)
+
+    ```plain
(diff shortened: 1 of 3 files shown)


src/actions/guides/getting-started/installing.cr
```diff
@@ -29,51 +27,127 @@ class Guides::GettingStarted::Installing < GuideAction
...
+    [Postgres CLI tools](https://postgresapp.com/documentation/cli-tools.html) are installed
+
+    ```plain
+    sudo mkdir -p /etc/paths.d &&
+      echo /Applications/Postgres.app/Contents/Versions/latest/bin | sudo tee /etc/paths.d/postgresapp
+    ```
+
+    There are other installation methods available in [Postgres CLI tools docs](https://postgresapp.com/documentation/cli-tools.html)
+
+    ## Linux dependencies
```
jwoertink: Are these for Crystal? Or things Lucky needs if you're running on Linux?

src/actions/guides/getting-started/installing.cr
```diff
@@ -29,51 +27,127 @@ class Guides::GettingStarted::Installing < GuideAction
...
+
+    Once the required dependencies above are installed, set up Lucky for your system.
+
+    ### Install Homebrew
+
+    Installation instructions from the [Homebrew website](https://brew.sh/)
+
+    ```plain
+    sudo mkdir -p /etc/paths.d &&
+    echo /Applications/Postgres.app/Contents/Versions/latest/bin | sudo tee /etc/paths.d/postgresapp
```
jwoertink: These are the wrong instructions for this section

src/actions/guides/getting-started/installing.cr
```diff
@@ -29,51 +27,127 @@ class Guides::GettingStarted::Installing < GuideAction
...
+    cd lucky_cli
+    ```
+
+    ### Check out the latest released version
+
+    ```plain
+    git checkout #{LuckyCliVersion.current_tag}
+    ```
+
+    > See a list of [released version](https://github.com/luckyframework/lucky_cli/releases)
```
jwoertink: Not sure why, but reading down this, this part threw me off. I thought that was the end of the instructions, and I thought "what about building the binary?". Then I went back and looked and went "oh, there's more".. Maybe just make this text inside of a side note?

jwoertink (changes requested):
Not sure why I didn't get notified about this before.. Or maybe I did and missed it? Overall this looks really nice. Only 1 major issue, but the rest looks decent.

src/actions/guides/getting-started/installing.cr
paulcsmith: Both I think...I don't know though..

I'm guessing these should be installed before anything else so I'll move it to the top

src/actions/guides/getting-started/installing.cr
paulcsmith: Oops :)

src/actions/guides/getting-started/installing.cr
paulcsmith: Yeah good idea! It's not that important so can be a regular text note

paulcsmith:
After going through this I realized I should probably flip this around a bit. Install Crystal and the CLI, then install yarn and all that since those are optional deps anyway. Will ping once that's ready. 

THanks for the review!

paulcsmith:
Ok I redid a bunch of stuff. LMK what you think

src/actions/guides/getting-started/installing.cr
```diff
@@ -29,54 +163,63 @@ class Guides::GettingStarted::Installing < GuideAction
...
+    There are other installation methods available in [Postgres CLI tools docs](https://postgresapp.com/documentation/cli-tools.html)
+
+    ### 2. Ensure Postgres CLI tools installed
+
+    ```plain
+    psql --version
+    ```
 
-    ### Dependencies for browser tests
+    Should return `psql (PostgreSQL) 12.x` or higher
```
jwoertink: Avram works on 9.6 or later, but maybe we just say 10.x or higher? 12 is really new, so some people may think "oh, I have to upgrade postgres to run this?"

jwoertink (approved):
This looks better 👍  I think it's good having that optional stuff at the bottom there. Plus at the top it makes it easier for people to skip to their setup right away.

src/actions/guides/getting-started/installing.cr
paulcsmith: Good point! I think 10.x sounds good

PR #258, opened 2020-03-31
Replace `Authentication` with `Authorization` header in CORS example code

David3141:
Using the standard `Authorization` header instead of `Authentication` in the example code might prevent errors for people using it in an API-only server (like me, lol)—especially since it's so far back in the line. If `Authentication` was intended: Sorry and feel free to just close this PR!


jwoertink (approved):
Good catch! I think this was my bad. I may have just typed it out without reading 😅  `Authorization` should be the right one here.

PR #259, opened 2020-04-01
Ducoument dynamic_asset

albertorestifo:
In the assets fronted guide the `dynamic_asset` function was not mentioned anywhere.

Given that I just stumbled on [this issue](https://github.com/luckyframework/lucky/issues/269) after encountering this problem, I thought it would be a good idea to add a line about it.


jwoertink (approved):
Oh nice catch! Yeah, I actually use this one all the time. Thanks for adding it in

paulcsmith (commented):
Thank you!

PR #265, opened 2020-04-07
Adding new reload model guide

### LINKED ISSUE #242: Document Avram model reload
https://github.com/luckyframework/avram/pull/324

jwoertink:
Fixes #242

This adds a new section for reloading to the query guide. It also mentions how to add in your preloads after you have your model. I've also added a small section to testing with an example of reloading your model after you've updated it during a spec.

### DIFF
--- a/src/actions/guides/database/querying-deleting.cr
+++ b/src/actions/guides/database/querying-deleting.cr
@@ -1,5 +1,6 @@
 class Guides::Database::QueryingDeleting < GuideAction
   ANCHOR_PRELOADING = "perma-preloading"
+  ANCHOR_RELOADING = "perma-reloading"
   guide_route "/database/querying-deleting"
 
   def self.title
@@ -556,6 +557,7 @@ class Guides::Database::QueryingDeleting < GuideAction
     task.user!
     ```
 
+    #{permalink(ANCHOR_RELOADING)}
     ## Reloading
 
     Reloading a model can be useful when you've loaded a model, but then there is a change to the data.
--- a/src/actions/guides/database/querying-deleting.cr
+++ b/src/actions/guides/database/querying-deleting.cr
@@ -556,6 +556,53 @@ class Guides::Database::QueryingDeleting < GuideAction
     task.user!
     ```
 
+    ## Reloading
+
+    Reloading a model can be useful when you've loaded a model, but then there is a change to the data.
+
+    ```crystal
+    author = AuthorQuery.find(5)
+
+    # Let's say the Author's profile picture is hidden
+    author.hide_face #=> true
+
+    # If this database value is updated...
+    SaveAuthor.update!(author, hide_face: false)
+
+    # We can reload to get the new value
+    author.reload.hide_face #=> false
+    ```
+
+    When calling the `reload` method on the [model](#{Guides::Database::Models.path}), the original
+    instance is not updated.
+
+    ```crystal
+    # The new value grabbed from the reloaded model
+    author.reload.hide_face #=> false
+
+    # The original value is still in place
+    author.hide_face #=> true
+    ```
+
+    ### Reloading with preloading
+
+    You can also use the `reload` method to preload associations. For example, if you
+    have a post, and want to get the comments preloaded, you can use `reload` with a block. This
+    will help you with performance by avoiding N+1 performance issues.
+
+    ```crystal
+    post = PostQuery.first
+
+    # This is not preloaded, and can lead to performance issues
+    post.comments
+
+    # Preload the comments for better performance
+    post.reload(&.preload_comments).comments
+    ```
+
+    Read up on [preloading associations](#{Guides::Database::QueryingDeleting.path(anchor: Guides::Database::QueryingDeleting::ANCHOR_PRELOADING)})
+    for more information.
+
     ## No results
 
     Avram gives you a `none` method to return no results. This can be helpful when under
--- a/src/actions/guides/testing/creating_test_data.cr
+++ b/src/actions/guides/testing/creating_test_data.cr
@@ -173,15 +173,16 @@ class Guides::Testing::CreatingTestData < GuideAction
 
       SavePost.update!(post, title: "New Post Title")
 
-      # We still have our original value
+      # The `post` still has the original data
       post.title.should eq "Custom Post"
 
-      # Now we have the new value
-      post.reload.title.should eq "New Post Title"
+      # Reloading returns a new model with the updated data
+      updated_post = post.reload
+      updated_post.title.should eq "New Post Title"
     end
     ```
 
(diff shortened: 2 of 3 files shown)


src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+    #{permalink(ANCHOR_RELOADING)}
+    ## Reloading
+
+    Reloading a model can be useful when you've loaded a model, but then there is a change to the data.
+
+    ```crystal
+    author = AuthorQuery.find(5)
+
+    # Let's say the Author's profile picture is hidden
+    author.hide_face #=> true
```
paulcsmith: LOL maybe `hide_avatar`? I keep imagining someone putting a bag over their face or something 🤣

src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+
+    ```crystal
+    # The new value grabbed from the reloaded model
+    author.reload.hide_face #=> false
+
+    # The original value is still in place
+    author.hide_face #=> true
+    ```
+
+    ### Reloading with preloading
```
paulcsmith: Maybe this?

```suggestion
    ### Adding preloads when reloading
```

src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+
+    You can also use the `reload` method to preload associations. For example, if you
+    have a post, and want to get the comments preloaded, you can use `reload` with a block. This
```
paulcsmith:
```suggestion
    have a post, and want to preload comments, you can use `reload` with a block. This
```

src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+    will help you with performance by avoiding N+1 performance issues.
```
paulcsmith: I think this line could be removed or reworded since I think the only reason to do this is for performance reasons. So maybe "This will help with performance by avoiding N+1" but maybe it would be better to remove it and link to the preloading docs? ANd put the N+1 stuff there?

src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+
+    ```crystal
+    post = PostQuery.first
+
+    # This is not preloaded, and can lead to performance issues
+    post.comments
```
paulcsmith: In this case preloading would be the same, might be better to show nested, like `post.comments.map(&.author.name) # Get all the author's names for this post`

src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+
+    # Preload the comments for better performance
+    post.reload(&.preload_comments).comments
```
paulcsmith:
```suggestion
    post.reload(&.preload_comments(PostQuery.preload_authors).comments.map(&.author.name)
```

src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+    ```
+
+    Read up on [preloading associations](#{Guides::Database::QueryingDeleting.path(anchor: Guides::Database::QueryingDeleting::ANCHOR_PRELOADING)})
+    for more information.
```
paulcsmith: Nice!

src/actions/guides/database/querying-deleting.cr
```diff
@@ -556,6 +557,54 @@ class Guides::Database::QueryingDeleting < GuideAction
...
+    ```
+
+    ### Reloading with preloading
+
+    You can also use the `reload` method to preload associations. For example, if you
+    have a post, and want to get the comments preloaded, you can use `reload` with a block. This
+    will help you with performance by avoiding N+1 performance issues.
+
+    ```crystal
+    post = PostQuery.first
```
paulcsmith: Maybe show creating a post? since with this example it'd instead be better to preload using the query. That way it saves a query

Might be worth noting that this should usually be used with a newly created/updated record since you can use preloads on the query directly in just about every other case

src/actions/guides/testing/creating_test_data.cr
```diff
@@ -162,6 +162,27 @@ class Guides::Testing::CreatingTestData < GuideAction
...
+
+    If you've made a change to your data, you can call `reload` to get the updated data.
+
+    ```crystal
+    it "updates a post title" do
+      post = PostBox.create &.title("Custom Post")
+
+      SavePost.update!(post, title: "New Post Title")
+
+      # We still have our original value
```
paulcsmith:
```suggestion
      # The 'post' still has the original data
```

src/actions/guides/testing/creating_test_data.cr
```diff
@@ -162,6 +162,27 @@ class Guides::Testing::CreatingTestData < GuideAction
...
+      post.title.should eq "Custom Post"
+
+      # Now we have the new value
```
paulcsmith:
```suggestion
      # Reloading returns a new model with the updated data
```

src/actions/guides/testing/creating_test_data.cr
```diff
@@ -162,6 +162,27 @@ class Guides::Testing::CreatingTestData < GuideAction
...
+      post.reload.title.should eq "New Post Title"
```
paulcsmith: One tiny suggestion. I think it would be good to reassign to a local var as it is fairly common to perform multiple assertions and reloading for each one would be pretty unperformant.

```crystal
updated_post = post.reload
updated_post.title.should eq "New Post Title"
```

paulcsmith (approved):
A few suggestions, but looks good to go when you're ready!

src/actions/guides/database/querying-deleting.cr
jwoertink: It might confusing people with a new record because there wouldn't be comments or authors just yet. Maybe I just do a comment saying something like

```crystal
# `post` is a recently updated record
```

paulcsmith: Oh yeah great point! I could see that being confusing. I like you comment idea :+1:

paulcsmith (approved):

PR #266, opened 2020-04-08
Added new guide on soft deletes

### LINKED ISSUE #245: Add soft delete docs
https://github.com/luckyframework/avram/pull/323

jwoertink:
Fixes #245

This PR moves the deleting guide out of querying and in to its own home. It also adds in the new Soft Delete API.

I decided to move it to its own page since the Query page is starting to get larger, and will soon be even larger with some more additions coming. Plus, this guide is pretty heft with the new soft deletes.

### DIFF
--- a/src/actions/guides/database/deleting_records.cr
+++ b/src/actions/guides/database/deleting_records.cr
@@ -143,7 +143,7 @@ class Guides::Database::DeletingRecords < GuideAction
     ### Default queries without soft deleted
 
     If you want to filter out soft deleted records by default, it's really easy to do.
-    Just add the `only_kept` method to your `initialize`.
+    Just add the `only_kept` method to an `initialize` method.
 
     ```crystal
     class ArticleQuery < Article::BaseQuery
--- a/src/actions/guides/database/deleting_records.cr
+++ b/src/actions/guides/database/deleting_records.cr
@@ -142,7 +142,7 @@ class Guides::Database::DeletingRecords < GuideAction
 
     ### Default queries without soft deleted
 
-    If you want to skip all of your soft deleted records by default, it's really easy to do.
+    If you want to filter out soft deleted records by default, it's really easy to do.
     Just add the `only_kept` method to your `initialize`.
 
     ```crystal
--- a/src/actions/guides/database/deleting_records.cr
+++ b/src/actions/guides/database/deleting_records.cr
@@ -11,8 +11,7 @@ class Guides::Database::DeletingRecords < GuideAction
 
     ### Delete one
 
-    Deleting a single record is actually done on the [model](#{Guides::Database::Models.path}) directly.
-    Since each query returns an instance of the model, you can just call `delete` on that record.
+    Deleting a single record is done on the [model](#{Guides::Database::Models.path}) directly.
 
     ```crystal
     user = UserQuery.find(4)
@@ -39,14 +38,14 @@ class Guides::Database::DeletingRecords < GuideAction
     Avram comes with some built-in modules to help working with soft deleted records a lot easier. Let's add it
     to an existing `Article` model.
 
-    * We need to add a new `soft_deleted_at : Time?` column to our table that we want soft deletes.
+    * First, we need to add a new `soft_deleted_at : Time?` column to the table that needs soft deletes.
 
     ```
     # Run this in your terminal
-    lucky gen.migration AddSoftDeleteToArticle
+    lucky gen.migration AddSoftDeleteToArticles
     ```
 
-    * Open your new `db/migrations/#{Time.utc.to_s("%Y%m%d%H%I%S")}_add_soft_delete_to_article.cr` file.
+    * Open your new `db/migrations/#{Time.utc.to_s("%Y%m%d%H%I%S")}_add_soft_delete_to_articles.cr` file.
 
     ```crystal
     def migrate
@@ -60,21 +59,23 @@ class Guides::Database::DeletingRecords < GuideAction
 
     ```crystal
     class Article < BaseModel
-      # Include this module
+      # Include this module to add methods for
+      # soft deleting and restoring
       include Avram::SoftDelete::Model
 
       table do
-        # Add this to your table
+        # Add the new column to your model
         column soft_deleted_at : Time?
       end
     end
     ```
 
-    * Next you just need to update `src/queries/article_query.cr`.
+    * Next you need to update `src/queries/article_query.cr`.
 
     ```crystal
     class ArticleQuery < Article::BaseQuery
-      # Add this module
+      # Include this module to add methods for
+      # querying and soft deleting records
       include Avram::SoftDelete::Query
     end
     ```
@@ -97,15 +98,14 @@ class Guides::Database::DeletingRecords < GuideAction
     article.reload.soft_deleted? #=> true
     ```
 
-    ### Marking all records as soft deleted
+    ### Soft deleting in bulk
 
-    You can bulk update a group of records as soft deleted with the `soft_delete` method called on a Query object
-    instead of the model directly.
+    You can bulk update a group of records as soft deleted with the `soft_delete` method on your Query object.
 
     ```crystal
     articles_to_delete = ArticleQuery.new.created_at.gt(3.years.ago)
 
-    # Marks only the articles created over 3 years ago as soft deleted
+    # Marks the articles created over 3 years ago as soft deleted
     articles_to_delete.soft_delete
     ```
 
@@ -118,15 +118,15 @@ class Guides::Database::DeletingRecords < GuideAction
     article.restore
     ```
 
-    ### Restoring all soft deleted records
+    ### Bulk restoring soft deleted records
 
-    The same as bulk updating records as soft deleted, we can also bulk update to restore them with
+    The same as we can bulk soft delet records, we can also bulk update to restore them with
     the `restore` method.
 
     ```crystal
-    articles_to_restore = ArticleQuery.new.only_soft_deleted
+    articles_to_restore = ArticleQuery.new.published_at.lt(1.week.ago)
 
-    # Restore all of the soft deleted records
+    # Restore recently published articles
     articles_to_restore.restore
     ```
 
@@ -138,8 +138,33 @@ class Guides::Database::DeletingRecords < GuideAction
 
     # Return all articles that are soft deleted
     ArticleQuery.new.only_soft_deleted
+    ```
 
-    # Return all articles whether soft deleted or not
+    ### Default queries without soft deleted
+
+    If you want to skip all of your soft deleted records by default, it's really easy to do.
+    Just add the `only_kept` method to your `initialize`.
+
+    ```crystal
+    class ArticleQuery < Article::BaseQuery
+      include Avram::SoftDelete::Query
+
+      # All queries will scope to only_kept
+      def initialize
+        only_kept
+      end
+    end
+    ```
+
+    ```crystal
+    # Return all articles that are not soft deleted
+    ArticleQuery.new
+    ```
+
+    Even with your default scope, you can still return soft deleted records when you need.
+
+    ```crystal
+    # Return all articles, both `kept` and soft deleted
     ArticleQuery.new.with_soft_deleted
     ```
 
--- a/src/actions/guides/database/deleting_records.cr
+++ b/src/actions/guides/database/deleting_records.cr
@@ -1,0 +1,169 @@
+class Guides::Database::DeletingRecords < GuideAction
+  guide_route "/database/deleting-records"
+
+  def self.title
+    "Deleting records"
+  end
+
+  def markdown : String
+    <<-MD
+    ## Deleting Records
+
+    ### Delete one
+
+    Deleting a single record is actually done on the [model](#{Guides::Database::Models.path}) directly.
+    Since each query returns an instance of the model, you can just call `delete` on that record.
+
+    ```crystal
+    user = UserQuery.find(4)
+
+    # DELETE FROM users WHERE users.id = 4
+    user.delete
+    ```
+
+    ### Bulk delete
+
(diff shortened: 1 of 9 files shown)


src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    # Run this in your terminal
+    lucky gen.migration AddSoftDeleteToArticle
+    ```
+
+    * Open your new `db/migrations/#{Time.utc.to_s("%Y%m%d%H%I%S")}_add_soft_delete_to_article.cr` file.
+
+    ```crystal
+    def migrate
+      alter table_for(Article) do
+        add soft_deleted_at : Time?
```
edwardloveall, line 53: I like that you called the column `soft_deleted_at` 👍

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    ### Query soft deleted records
+
+    ```crystal
+    # Return all articles that are not soft deleted
+    ArticleQuery.new.only_kept
+
+    # Return all articles that are soft deleted
+    ArticleQuery.new.only_soft_deleted
+
+    # Return all articles whether soft deleted or not
```
edwardloveall: The phrasing here is a little confusing. Maybe
> Return all articles, both `kept` and `soft_deleted`.

or something. I think it would also be worth saying that the normal `.all` now only returns `kept` records so there's a point of comparison.

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    ### Truncate database
+
+    You can also truncate your entire database by calling `truncate` on your database class.
+
+    ```crystal
+    AppDatabase.truncate
+    ```
+
+    > This method is great for tests; horrible for production. Also note this method is not chainable.
```
edwardloveall, line 191: An idle thought: what if we made another method that checked for the production environment and didn't allow you to call `truncate` on AppDatabase unless it was like `AppDatabase.production_truncate` or `.truncate.confirm`.

edwardloveall (approved):
This is a great guide. Thanks for writing it up.

src/actions/guides/database/deleting_records.cr
jwoertink: That was definitely a @paulcsmith choice, but I like it too. 😄

src/actions/guides/database/deleting_records.cr
jwoertink: Is that what happens? That's what I was thinking, but couldn't find in the code where `.all` does a default scope.. I agree with the wording change. I'm horrible at that 😅

src/actions/guides/database/deleting_records.cr
jwoertink: We do have `Lucky::Env.production?`, so that's probably not a bad idea.

paulcsmith:
I'm not sure the querying guides appreciate being called "hefty", Jeremy

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
+class Guides::Database::DeletingRecords < GuideAction
```
paulcsmith, line 1: Huge +1 on extracting this. Good idea!

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    "Deleting records"
+  end
+
+  def markdown : String
+    <<-MD
+    ## Deleting Records
+
+    ### Delete one
+
+    Deleting a single record is actually done on the [model](#{Guides::Database::Models.path}) directly.
```
paulcsmith:
```suggestion
    Deleting a single record is done on the [model](#{Guides::Database::Models.path}) directly.
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    Since each query returns an instance of the model, you can just call `delete` on that record.
```
paulcsmith: This isn't quite right since queries can return a Queryable and not a collection. I think it may be more clear to leave this off and show only the example. I think it is pretty clear like that

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    ```
+
+    ### Bulk delete
+
+    If you need to bulk delete a group of records based on a where query, you can use `delete` at
+    the end of your query. This returns the number of records deleted.
+
+    ```crystal
+    # DELETE FROM users WHERE banned_at IS NOT NULL
+    UserQuery.new.banned_at.is_not_nil.delete
```
paulcsmith, line 30: You know what, this gives me an idea for adding `is_present`. The `isn_not_nil` always trips me up. I'm gonna open an issue for that https://github.com/luckyframework/avram/issues/344

But this is good to go as-is. Just a thought as I was reading through it

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    ## Soft Deletes
+
+    A "soft delete" is when you want to hide a record as if it were deleted, but you want to keep the actual
+    record in your database. This allows you to restore the record without losing any previous data or associations.
+
+    Avram comes with some built-in modules to help working with soft deleted records a lot easier. Let's add it
+    to an existing `Article` model.
+
+    * We need to add a new `soft_deleted_at : Time?` column to our table that we want soft deletes.
```
paulcsmith:
```suggestion
    * First, we need to add a new `soft_deleted_at : Time?` column to the table that needs soft deletes.
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    ```
+    # Run this in your terminal
+    lucky gen.migration AddSoftDeleteToArticle
```
paulcsmith:
```suggestion
    lucky gen.migration AddSoftDeleteToArticles
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    * Now open your `src/models/article.cr` file.
+
+    ```crystal
+    class Article < BaseModel
+      # Include this module
+      include Avram::SoftDelete::Model
+
+      table do
+        # Add this to your table
```
paulcsmith:
```suggestion
        # Add the new column to your model
```

Since it is kind of already in the db table. So maybe this is clearer? Up to you

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+        column soft_deleted_at : Time?
+      end
+    end
+    ```
+
+    * Next you just need to update `src/queries/article_query.cr`.
```
paulcsmith:
```suggestion
    * Next you need to update `src/queries/article_query.cr`.
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    ```crystal
+    class ArticleQuery < Article::BaseQuery
+      # Add this module
```
paulcsmith: I think it'd be cool to give a brief description of what it adds. Something like this maybe?

```suggestion
      # Include this module to add methods for querying soft deleted records
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+        add soft_deleted_at : Time?
+      end
+    end
+    ```
+
+    * Now open your `src/models/article.cr` file.
+
+    ```crystal
+    class Article < BaseModel
+      # Include this module
```
paulcsmith:
```suggestion
      # Include this module to add methods for soft deleting and restoring
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    article.soft_deleted? #=> false
+
+    # Save the record as soft deleted
+    article.soft_delete
+
+    # Reload the model to get the new value
+    article.reload.soft_deleted? #=> true
+    ```
+
+    ### Marking all records as soft deleted
```
paulcsmith:
```suggestion
    ### Soft deleting multiple records at once
```

paulcsmith: Or maybe "Soft deleting in bulk"?

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    You can bulk update a group of records as soft deleted with the `soft_delete` method called on a Query object
```
paulcsmith:
```suggestion
    You can bulk update a group of records as soft deleted with the `soft_delete` method on your Query object
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    instead of the model directly.
```
paulcsmith: I think this part can be omitted but up to you!

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    ```crystal
+    articles_to_delete = ArticleQuery.new.created_at.gt(3.years.ago)
+
+    # Marks only the articles created over 3 years ago as soft deleted
```
paulcsmith:
```suggestion
    # Marks the articles created over 3 years ago as soft deleted
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    ### Restore a soft deleted record
+
+    If you need to restore a soft deleted record, you can use the `restore` method.
+
+    ```crystal
+    # Set the `soft_deleted_at` back to `nil`
+    article.restore
+    ```
+
+    ### Restoring all soft deleted records
```
paulcsmith:
```suggestion
    ### Bulk restoring soft deleted records
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    The same as bulk updating records as soft deleted, we can also bulk update to restore them with
```
paulcsmith:
```suggestion
    The same as we can bulk  soft delete records, we can also bulk restore them with
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+    the `restore` method.
+
+    ```crystal
+    articles_to_restore = ArticleQuery.new.only_soft_deleted
```
paulcsmith: Since the `only_soft_deleted` scope has not been introduced, maybe a different query would be clearer?

```suggestion
    articles_to_restore = ArticleQuery.new.published_at.gt(1.week.ago)
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,169 @@
...
+
+    # Restore all of the soft deleted records
```
paulcsmith:
```suggestion
    # Restore recently updated articles
```

src/actions/guides/database/deleting_records.cr
paulcsmith: I don't think `all` does have a default scope. That was done on purpose so the module doesn't start changing defaults unless you do it. We should document how to do that though.

You add `only_kept` to `def initialize` and then your queries will default to only_kept. Then you can include everything with `with_soft_deleted` https://github.com/luckyframework/avram/blob/3b8c2a38e1805f7f3b26fc0e82bddf8d6084dcdf/src/avram/soft_delete/query.cr#L34-L36

paulcsmith: So maybe this particular part should go in a different section? Like `Only show kept records by default` that shows how to make that behavior the default and how to include soft deleted if you want

src/actions/guides/database/deleting_records.cr
paulcsmith: Yeah we should definitely do that. Rails does something similar. I think this could be a setting in Habitat. `allow_truncate` then in the default config we do `settings.allow_truncate = !Lucky::Env.production?`

Since we can't rely on Lucky::Env in Avram (Lucky is not required for Avram)

paulcsmith (approved):
Fantastic set of changes. I like that it is split into its own thing. Soft deleted is also very clear. Love it.

Just a few wording suggestions and clarification on how to show only kept by default.

jwoertink:
Awesome! Thanks for the review @edwardloveall and @paulcsmith! And I'm sorry Query Guides, you're not hefty, just full featured 😘 😄

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,194 @@
...
+    # Return all articles that are not soft deleted
+    ArticleQuery.new.only_kept
+
+    # Return all articles that are soft deleted
+    ArticleQuery.new.only_soft_deleted
+    ```
+
+    ### Default queries without soft deleted
+
+    If you want to skip all of your soft deleted records by default, it's really easy to do.
```
paulcsmith:
```suggestion
    If you want to filter out soft deleted records by default, it's really easy to do.
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -0,0 +1,194 @@
...
+    Just add the `only_kept` method to your `initialize`.
```
paulcsmith:
```suggestion
    Just add the `only_kept` method to an `initialize` method.
```

Changed this since by default there isn't one set so might be confusing.

paulcsmith (approved):
2 tiny suggestions. Looks good.

And mega lol at our "full-featured" guides 😂

PR #268, opened 2020-04-08
Remove crenv recommendation

paulcsmith:
crenv is not actively maintained, and currently fails to install 0.34

asdf is more stable the asdf-crystal is part of asdf-community which has
more reviewers and maintainers.

### DIFF
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -27,16 +27,22 @@ class Guides::GettingStarted::Installing < GuideAction
 
     ### 3. Configure SSL for Crystal
 
-    You'll need to add `export PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig`
-    to your `~/.bash_profile` or `~/.zshrc` so Crystal can use SSL through the
-    openssl package.
+    You'll need to add tell Crystal how to find OpenSSL by adding an `export`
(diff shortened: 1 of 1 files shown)


src/actions/guides/getting-started/installing.cr
```diff
@@ -31,11 +31,15 @@ class Guides::GettingStarted::Installing < GuideAction
     to your `~/.bash_profile` or `~/.zshrc` so Crystal can use SSL through the
     openssl package.
 
-    ```shell
-    # For Bash, the default shell on MacOS
+    **For Bash, the default shell on MacOS:**
```
jwoertink: Default is now `zsh` if you have a new mac with Catalina. I think it's only bash if you upgraded to Catalina. [ref](https://www.theverge.com/2019/6/4/18651872/apple-macos-catalina-zsh-bash-shell-replacement-features)

jwoertink: Maybe just say `For Bash` and `For ZSH`, and then a note that if you're unsure, you can `echo $SHELL` ?

jwoertink (approved):
This looks good. 1 small note, but otherwise good to go!

src/actions/guides/getting-started/installing.cr
paulcsmith: This is great to know! I’ll definitely change this and mention the echo $SHELL to determine which one

PR #269, opened 2020-04-09
Updates/lucky0.20

jwoertink:
This updates the website to Crystal 0.34 and Lucky 0.20. I've also made some of the optional updates like using the getter methods from `needs`, and added in the new system_check stuff. 

The only warning left is related to Dexter, but we already know about that.


paulcsmith (approved):
Nice!! LGTM

Just out of curiosity, how long did this take? Was it fairly painless?

jwoertink:
Maybe like 15min? It wasn't long at all.

PR #270, opened 2020-04-10
Adding some missing query guides

### LINKED ISSUE #237: Document missing query methods
I found that `group` is missing, and I'm sure there's some others missing as well. We should go through all the things a query object can do, and make sure they're all documented.

jwoertink:
Fixes #237


paulcsmith (approved):
LGTM!

PR #272, opened 2020-04-10
Update page guides with new getter from needs

### LINKED ISSUE #233: Modify guide to use getters instead of instance vars in pages
https://github.com/luckyframework/lucky/pull/1034

Note that guide fixes should not be merged until a new version is released

### LINKED ISSUE #250: Document that `needs` in pages sets up a getter
And update examples to use the getter

jwoertink:
Fixes #250 
Fixes #233

### DIFF
--- a/src/actions/guides/frontend/rendering_html.cr
+++ b/src/actions/guides/frontend/rendering_html.cr
@@ -63,8 +63,8 @@ class Guides::Frontend::RenderingHtml < GuideAction
     ## Declaring what a page needs
 
     You’ll notice we used `needs` near the top of the class. This declares that for
-    this page to render we need an Array of Strings and that they will be assigned
-    to the `user_names` getter method. We set the user names by passing it in the
+    this page to render we need an Array of Strings and that they will be accessible
+    from the `user_names` getter method. We set the user names by passing it in the
     `html` macro in our action: `html IndexPage, user_names: ["Paul", "Sally", "Jane"]`
 
     > This is nice because you won’t accidentally forget to pass something to a page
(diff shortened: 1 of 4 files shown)


src/actions/guides/frontend/rendering_html.cr
```diff
@@ -64,7 +64,7 @@ class Guides::Frontend::RenderingHtml < GuideAction
 
     You’ll notice we used `needs` near the top of the class. This declares that for
     this page to render we need an Array of Strings and that they will be assigned
-    to the `@user_names` variable. We set the user names by passing it in the
+    to the `user_names` getter method. We set the user names by passing it in the
```
paulcsmith: Maybe:

```suggestion
    to the `@user_names` instance variable that is accessible with a getter method. We set the user names by passing it in the
```

A little longer but may be clearer. Thoughts on that?

paulcsmith (approved):
Looks great!

src/actions/guides/frontend/rendering_html.cr
jwoertink: I think it might be a little confusing that we show the instance variable, but then say don't use it, or show it being used. If you have a strong understanding of Crystal, then it makes sense, but I think if you're new to both Crystal and Lucky, then you might be a little confused on the instance variable. So not too sure on this.

paulcsmith: Yeah that is a really good point. Maybe what was confusing me was the "assigned to a getter". Maybe "is accessible from a getter"?

jwoertink: Ah, gotcha. That makes sense.

PR #274, opened 2020-04-10
Added a little clarity for the heroku deploy guide

### LINKED ISSUE #256: Update heroku deploy guide to mention API
If you generate an API app, then add the nodejs heroku buildpack, your deployment will fail. We should mention that it's only for the web apps. 😅  oops..

https://luckyframework.org/guides/deploying/heroku

jwoertink:
Fixes #256 

Also fixed a small css issue while I was in there:
Before:
Screen Shot 2020-04-10 at 1 18 33 PM

After:
Screen Shot 2020-04-10 at 1 19 28 PM

### DIFF
--- a/src/actions/guides/deploying/heroku.cr
+++ b/src/actions/guides/deploying/heroku.cr
@@ -21,8 +21,15 @@ class Guides::Deploying::Heroku < GuideAction
     `heroku git:remote -a <APP-NAME>`
     * Add the following buildpacks in
     order:
-      * `heroku buildpacks:add https://github.com/heroku/heroku-buildpack-nodejs`
-      * `heroku buildpacks:add https://github.com/luckyframework/heroku-buildpack-crystal`
+
+    ```plain
+    # Add this if your app has HTML and Assets
+    heroku buildpacks:add https://github.com/heroku/heroku-buildpack-nodejs
+
+    # Then add this for all Lucky apps
+    heroku buildpacks:add https://github.com/luckyframework/heroku-buildpack-crystal
+    ```
+
     * Set `LUCKY_ENV` to `production`:
     `heroku config:set LUCKY_ENV=production`
     * Set `SECRET_KEY_BASE`:
@@ -89,8 +96,7 @@ class Guides::Deploying::Heroku < GuideAction
     to our app:
 
     ```bash
-    $ heroku buildpacks:add
-    https://github.com/heroku/heroku-buildpack-nodejs
+    $ heroku buildpacks:add https://github.com/heroku/heroku-buildpack-nodejs
     ```
 
     While Heroku has a buildpack for Node.js apps,
@@ -103,8 +109,7 @@ class Guides::Deploying::Heroku < GuideAction
     and install the dependencies for Lucky:
 
     ```bash
-    $ heroku buildpacks:add
-    https://github.com/luckyframework/heroku-buildpack-crystal
+    $ heroku buildpacks:add https://github.com/luckyframework/heroku-buildpack-crystal
     ```
 
     ## Adding environment variables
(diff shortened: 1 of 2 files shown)


src/actions/guides/deploying/heroku.cr
```diff
@@ -89,8 +96,7 @@ class Guides::Deploying::Heroku < GuideAction
     to our app:
 
     ```bash
-    $ heroku buildpacks:add
-    https://github.com/heroku/heroku-buildpack-nodejs
+    $ heroku buildpacks:add https://github.com/heroku/heroku-buildpack-nodejs
```
jwoertink: Made this 1 line so it can be copied as is.

src/actions/guides/deploying/heroku.cr
```diff
@@ -103,8 +109,7 @@ class Guides::Deploying::Heroku < GuideAction
     and install the dependencies for Lucky:
 
     ```bash
-    $ heroku buildpacks:add
-    https://github.com/luckyframework/heroku-buildpack-crystal
+    $ heroku buildpacks:add https://github.com/luckyframework/heroku-buildpack-crystal
```
jwoertink: ditto

src/css/components/_markdown-content.scss
```diff
@@ -113,14 +113,22 @@
...
 
     a:hover {
       @apply text-grey-lighter border-grey;
     }
   }
 
+  pre {
+    code {
+      a {
+        @apply no-underline text-grey border-grey-dark;
```
jwoertink, line 127: Keep the links in the big code blocks as they were before.

src/actions/guides/deploying/heroku.cr
```diff
@@ -21,8 +21,15 @@ class Guides::Deploying::Heroku < GuideAction
     `heroku git:remote -a <APP-NAME>`
     * Add the following buildpacks in
     order:
-      * `heroku buildpacks:add https://github.com/heroku/heroku-buildpack-nodejs`
-      * `heroku buildpacks:add https://github.com/luckyframework/heroku-buildpack-crystal`
+
+    ```plain
+    # Add this if your app has HTML and Assets
```
paulcsmith:
```suggestion
    # Skip this for API only apps. Add this if your app has HTML and Assets
```

Trying to make it a bit clearer when to skip. What do you think?

src/actions/guides/deploying/heroku.cr
paulcsmith: Good idea. Also I don't think we need the `$`. I've always found that to be unhelpful since sometimes it makes it harder to copy. Up to you though. Not a huge deal either way :P

src/css/components/_markdown-content.scss
paulcsmith: Love it! Thanks for the explanation

paulcsmith (approved):
Great changes! Thoughts on doing something similar to the install guide where we use `###` for each step and black code blocks with triple backtcks so they all stand out and look the same?

Something like:

```
### Add buildpacks

Add the following...

### Set env key

etc. with  ` ` ` backticks for code blocks
```

src/actions/guides/deploying/heroku.cr
jwoertink: I originally was going to remove it, but I saw we do it in so many places. I'll remove them all from this guide, and we can go through the other later to make it consistent.

jwoertink:
I think I know what you mean, but just to be clear, you're saying under the Quickstart guide, instead of doing a UL list, do it more like how we have the install guide at the beginning of the guides?

paulcsmith:
@jwoertink Yeah exactly. So it uses the smaller headers `###` for each step instead of a list item

PR #276, opened 2020-04-10
fixing code example in migrations. 

### LINKED ISSUE #196: Fix add column example
**Gitter:** https://gitter.im/luckyframework/Lobby?at=5de4589a46397c721c6d10a2

```crystal
def migrate
  alter table_for(User) do
    # Add nullable column first
    add otp_code : String?
  end

  # Then add values to it
  UserQuery.new.each do |user|
    User::SaveOperation.udpate!(user, otp_code: CodeGenerator.generate)
  end

  # Then make it non-nullable
  make_required :otp_code
end
```

that code example has at least two mistakes, one is the udpate, the other that make_required needs the table name as argument as well
from https://luckyframework.org/guides/database/migrations#using-fill_existing_with

jwoertink:
Fixes #196


paulcsmith (approved):
Nice!

PR #281, opened 2020-04-11
Add pagination guide

### LINKED ISSUE #240: Document Lucky::Paginator and components
https://github.com/luckyframework/lucky/pull/1020

Don't merge/deploy until next version of Lucky is released

paulcsmith:
Closes #240

### DIFF
--- a/src/actions/guides/database/pagination.cr
+++ b/src/actions/guides/database/pagination.cr
@@ -1,0 +1,422 @@
+class Guides::Database::Pagination < GuideAction
+  guide_route "/database/pagination"
+
+  def self.title
+    "Pagination"
+  end
+
+  def markdown : String
+    <<-MD
+    ## Introduction
+
+    In many frameworks pagination can be painful. You may have multiple
+    options to choose from, the HTML can be hard to customize, and often the
+    APIs are hard to understand.
+
+    In Lucky and Avram, we've built in pagination and HTML pagination
+    components so you can get started quickly. You can also copy the HTML
+    components to your app and customize them to your hearts content.
+
+    Let's dive in!
+
+    ## Setup
+
+    You can paginate records from within an action by including the
+    [`Lucky::Paginator::BackendHelpers`](https://github.com/luckyframework/lucky/blob/master/src/lucky/paginator/backend_helpers.cr)
+    in your actions.
+
+    It's usually best to include the helpers in your base actions, such as `ApiAction`
+    or `BrowserAction`. That way any action that inherits from them has access to the included methods:
+
+    ```crystal
+    # src/actions/browser_action.cr
+    abstract class BrowserAction < Lucky::Action
+      include Lucky::Paginator::BackendHelpers
+    end
+    ```
+
+    > We are using `BrowserAction` in this example, but it works the same for
+    ApiAction > or any other `Lucky::Action`
+
+    This module will add the `paginate` method to you actions. This is what
+    you'll use to paginate your query results.
+
+    ## Paginating
+
+    Pagination is easy-peasy, call `paginate` with a query:
+
+    ```crystal
+    class Users::Index < BrowserAction
+      get "/users" do
+        pages, users = paginate(UserQuery.new)
+        html IndexPages, users: users, pages: pages
+      end
+    end
+    ```
+
+    In this case `pages` is a `Lucky::Paginator` with metadata about the
+    current page, total numbers of pages, and helpers for generating paths to
+    next and previous pages.
+
+    The other variable, `users`, is the passed in query (`UserQuery.new`)
+    with a `limit` and `offset` applied so it returns records for just
+    the requested page.
+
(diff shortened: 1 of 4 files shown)


src/actions/guides/database/pagination.cr
```diff
@@ -0,0 +1,422 @@
+class Guides::Database::Pagination < GuideAction
```
jwoertink, line 1: Ah interesting. I was wondering how this would work since it affects both HTML and the Database.

src/actions/guides/database/pagination.cr
```diff
@@ -0,0 +1,422 @@
...
+      end
+    end
+    ```
+
+    In this case `pages` is a `Lucky::Paginator` with metadata about the
+    current page, total numbers of pages, and helpers for generating paths to
+    next and previous pages.
+
+    The other variable, `users`, is the passed in query (`UserQuery.new`)
+    with a `limit` and `offset` applied so it returns records for just
```
jwoertink, line 62: Oh cool! It already applies that.

src/actions/guides/database/pagination.cr
```diff
@@ -0,0 +1,422 @@
...
+
+    If this logic is not what you want, you can customize it by creating
+    a `paginator_page` that returns an `Int32`
+
+    ```crystal
+    abstract class ApiAction < Lucky::Action
+      include Lucky::Paginator::BackendHelpers
+
+      def paginator_page : Int32
+        request.headers["Page"]? || 1 # Use a header
```
jwoertink: Would `request.headers["Page"]` return a String?

src/actions/guides/database/pagination.cr
```diff
@@ -0,0 +1,422 @@
...
+
+    With just 2 lines of code you can render pagination links in your HTML pages.
+
+    * Add a `needs pages : Lucky::Paginator`
+    * Mount one of the built-in pagination components
+
+    ```crystal
+    class Users::IndexPage < MainLayout
+      needs users : UserQuery
+      needs pages : Lucky::Paginator # Add this to use the paginator
```
jwoertink, line 136: Side note, I wonder if we had a `needs_paginated users : UserQuery` that does both of these would be helpful 🤔

jwoertink (approved):
Awesome! This looks great. You can probably merge in, but I'm gonna do another pass once I'm done with a few things just in case.

jwoertink:
oh, actually, spoke too soon 😅  I think this should be updated too https://lucky-websit-pcs-240-pa-ljxzhi.herokuapp.com/guides/database/querying-records#pagination

src/actions/guides/database/pagination.cr
```diff
@@ -0,0 +1,422 @@
...
+    ```crystal
+    range = pages.item_range
+    "Showing #\{range.begin}-#\{range.end} of #\{pages.item_count}"
+    ```
+
+    ### `series`
+
+    Requires its own section 😃
+
+    ## The 'series' method
```
jwoertink, line 309: These quotes mess up the anchor. This is probably a bug in how these are generated. 

https://lucky-websit-pcs-240-pa-ljxzhi.herokuapp.com/guides/database/pagination#the-'series'-method

src/actions/guides/database/pagination.cr
paulcsmith: Yeah I wasn't sure what to do but decided to do it here. I was thinking of splitting it but that seems even worse. I figure people can find it with search on the site or Google.

src/actions/guides/database/pagination.cr
paulcsmith: Whoops! Yeah need to to_i it

src/actions/guides/database/pagination.cr
paulcsmith: I thought about it but it is really hard to do with the type system and also makes it a bit confusing because you're never sure if it is paginated or not :(

You could do a generic `Paginator(T)` with a query or whatever but then I think that gets fairly confusing. The other plus side of this is we can use the Paginator for arrays, or any enumberable at some point.

With that said I am open to changing this if we come up with something!

src/actions/guides/database/pagination.cr
paulcsmith: Ah ok. I'll leave this as-is and do a followup PR that slugifies these correctly!

paulcsmith (commented):
Thanks for the feedback. I changed the 2 spots where I forgot to `to_i` it!

paulcsmith:
Oh nice catch on pagination! Since this new section is quite large I think I'll delete the old one and leave this as its own thing. Kinda like we did for delete. I think that might be good since pagination has the html and stuff too.

Maybe we'll figure out a better spot for it at some point, but I think this will be ok...?

src/actions/guides/database/pagination.cr
jwoertink: I think having it here is fine. We can always revisit it later if need be.

PR #285, opened 2020-04-19
Update logger and bump version to 0.21.0

paulcsmith:
We need upgrade notes and changelog, but I'd like to get this out for people so new projects don't get the Dexter warning


paulcsmith:
Feel free to tweak or leave comments, but I'm pretty done with logging for a bit so I'd like to get this out there so we can do some other cool stuff :D

PR #290, opened 2020-04-26
Fix select fields documentation

matthewmcgarvey:
The options are determined by the passed in array of options. There were two more options in the HTML than what were in the Crystal code.


paulcsmith (approved):
Thank you for catching that and submitting a Pr

PR #292, opened 2020-04-26
You use group() not group_by() because the latter is a method on enum…

jwoertink:
…erable or whatever. That was a typo... my bad...


paulcsmith (approved):
Looks good!

PR #301, opened 2020-05-05
Update Guides with dynamic Crystal Version

jwoertink:
The Crystal version is used in a few spots for the installing guide. Moved this value to a method so it's easier to update later.


paulcsmith (approved):
YAS. Great idea

PR #306, opened 2020-05-15
Fix minor typo

UnsolvedCypher:
Change "invalid Operation's" to "invalid operations"


jwoertink (approved):
Good catch! Thanks for contributing!

PR #310, opened 2020-05-17
Specify environment vars to migrate

edwardloveall:
When migrating on Ubuntu, crystal and your app may need extra context to migrate the database. To specify that context, you can add the environment variables inline.

### DIFF
--- a/src/actions/guides/deploying/ubuntu.cr
+++ b/src/actions/guides/deploying/ubuntu.cr
@@ -124,6 +124,15 @@ class Guides::Deploying::Ubuntu < GuideAction
     crystal run tasks.cr -- db.migrate
     ```
 
+    > Note: You may have to specify the `DATABASE_URL` and any other evironment
+    > variables your app needs to migrate. Only `DATABASE_URL` needs to be
+    > real. The rest can be blank. You can add them before running crystal like
+    > so:
+    >
+    > ```bash
+    > API_KEY= SUPPORT_EMAIL= DATABASE_URL=postgresql://<username>:<password>@127.0.0.1/<appname>_production crystal run tasks.cr -- db.migrate
+    >```
+
     Exit your session as the `deploy` user, either
     with `CTRL-D` or by entering `exit`.
 


jwoertink (approved):

src/actions/guides/deploying/ubuntu.cr
```diff
@@ -124,6 +124,15 @@ class Guides::Deploying::Ubuntu < GuideAction
     crystal run tasks.cr -- db.migrate
     ```
 
+    > Note: You may have to specify the `DATABASE_URL` and any other evironment
+    > variables your app needs to migrate. Only `DATABASE_URL` needs to be
+    > real. The rest can be blank. You can add them before running crystal like
+    > so:
+    >
+    > ```bash
+    > API_KEY= SUPPORT_EMAIL= DATABASE_URL=postgresql://<username>:<password>@127.0.0.1/<appname>_production crystal run tasks.cr -- db.migrate
```
paulcsmith, line 133: I'm a bit confused by the SUPPORT_EMAIL and API_KEY. Are those just examples?

Maybe a good example would be `SENGRID_API_KEY`?

paulcsmith (commented):
Nice. This'll help a lot I think. When I first started I had no idea how to use ENV vars

src/actions/guides/deploying/ubuntu.cr
edwardloveall: Yup they are examples of custom environment variables in an app. Curiously I do have `SENGRID_API_KEY` which is required for app boot, but not migration! I can't think of any other built in ENV vars that would make better examples, but I'm all ears.

paulcsmith: Maybe PORT? This isn't a big deal though so don't feel too much pressure. I think this was an awesome change even left as-is.

Also just FYI. I've been a bit anxious/stressed and am sometimes being overly abrupt in my communication which can come off demanding. So if I do that, I'm sorry in advance and am working on it, but LMK because I definitely don't mean that

edwardloveall: I had the same deal with `PORT` as with `SENGRID_API_KEY`. What if I add a little more to the explanation like
> For example, if your app needs both the `API_KEY` and `SUPPORT_EMAIL` environment variable:

Sorry to hear that you're anxious/stressed. A lot of that going around these days. Your comments were A-OK as as I'm concerned!

paulcsmith: Yeah I like the idea of explaining. That should do the trick!

> Sorry to hear that you're anxious/stressed

Thanks ❤️

> our comments were A-OK as as I'm concerned!

Phew!

edwardloveall: https://github.com/luckyframework/website/pull/314

PR #311, opened 2020-05-18
Fix typo.

oneiros:
Great addition btw, @edwardloveall !


edwardloveall (approved):
Hah thanks @oneiros. I kept spelling this wrong in some of my server configs too which messed me up.

PR #315, opened 2020-05-20
Add Dokku deployment instructions

acoustep:


src/actions/guides/deploying/dokku.cr
```diff
@@ -0,0 +1,99 @@
+class Guides::Deploying::Dokku < GuideAction
+  guide_route "/deploying/dokku"
+
+  def self.title
+    "Deploying to Dokku"
```
paulcsmith, line 5:
```suggestion
    "Deploying with Dokku"
```

src/actions/guides/deploying/dokku.cr
```diff
@@ -0,0 +1,99 @@
+class Guides::Deploying::Dokku < GuideAction
```
paulcsmith, line 1: Could you run `crystal tool format src` to fix formatting? It looks like there is a formatting issue somewhere in here according to CI

paulcsmith (approved):
One tiny issue with formatting, other than that this looks great. Thank you so much!!

src/actions/guides/deploying/dokku.cr
paulcsmith: Never mind. The existing docs use `to` :)

src/actions/guides/deploying/dokku.cr
acoustep: Done :)

jwoertink (approved):
Nice! It's pretty cool having all these different options to give to people. I've never used dokku, but if this works as is, then I think it's good!

paulcsmith:
🥳 🚀

PR #316, opened 2020-06-02
Add kindmetrics

paulcsmith:


edwardloveall (approved):
🎉

stephendolan:
@paulcsmith do we still want to add this to the site now that Kindmetrics is out of their free beta?

paulcsmith:
Probably not. Also because I've been realizing metrics really make me anxious. I think I'd rather make stuff and not worry so much about how many people are looking at it!

PR #340, opened 2020-06-29
Blog post on 0.23 release

jwoertink:
This is a post announcing the 0.23.0 release.


src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+  def published_at : Time
+    Time.utc(year: 2020, month: 6, day: 30)
+  end
+
+  def summary : String
+    <<-TEXT
+    Lucky v0.23 is out now. Lots of new features, bug fixes,
+    and even some performance boosts!
+    TEXT
```
paulcsmith: What do you think about listing some of the improvements so it stands out a bit more?

Maybe something like: "Improved memoization, more powerful browser tests, routing speed improvements, and more!"

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    Be sure to upgrade your version of Crystal, and take a look at our
+    [UPGRADE NOTES](https://github.com/luckyframework/lucky/blob/master/UPGRADE_NOTES.md#upgrading-from-022-to-023)
+    for help with migrating your app.
+
+    ### LuckyDiff
+
+    Thanks to a community member [@stephendolan](https://github.com/stephendolan/lucky_diff) for
+    creating this awesome tool, we have started linking to it in our UPGRADE_NOTES to help you
+    see what the differences are between versions!
```
paulcsmith:
```suggestion
    ### How to upgrade
    
    See the [UPGRADE_NOTES](https://github.com/luckyframework/lucky/blob/master/UPGRADE_NOTES.md#upgrading-from-022-to-023) to learn how to upgrade to Lucky 0.23.0.

    Special thanks to community member [@stephendolan](https://github.com/stephendolan) for
    creating [LuckyDiff](https://luckydiff.com). This tool helps see what has changed between versions and will be included in upgrade notes from now on. 
```

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    ### What happened to 0.22.0?
+
+    Don't worry, you didn't miss anything. If you're using Lucky 0.21.0, you can upgrade to
+    [0.22.0](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md#v0220-2020-06-17) just by
+    updating your Crystal version to 0.35.0. This release didn't change anything other than compatibility
+    with the latest Crystal. To get all of the latest goodies, you'll want to go up to 0.23.0.
+
+    ## Quick Updates
```
paulcsmith: What do you think about this?

```suggestion
    ## Here's what's new
```

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    ```crystal
+    # Before
+    mount MyComponent.new(arg: 1)
+
+    # After
+    m MyComponent, arg: 1
+    ```
+
+    As you can see, you still have all the flexibility, but in a bit more compact version!
```
paulcsmith:
```suggestion
    You still have all the flexibility, but in a more compact version. The new `m` method also opens the doors for more functionality that we'll be adding to components in future releases of Lucky.
```

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    ### Params
```
paulcsmith:
```suggestion
    ### Trim extra whitespace in params
```

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    When you need to access a param value, you'll use `params.get(:key)`, but now, the blankspace
+    will be stripped automatically. If you need to get the untouched value, you can use `params.get_raw`.
```
paulcsmith:
```suggestion
    Leading and trailing whitespace is now stripped automatically when accessing params. If you need to get the unstripped value, you can use `params.get_raw(:the_key)`.
```

And maybe an example like `For example if a user sends "email@example.com " Lucky will remove the trailing whitespace and return `"email@example.com"`

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    ### Memoize 2.0
```
paulcsmith: To me this looks like there is a memoize shard with an actual 2.0. Maybe this could work?

```suggestion
    ### Revamped memoization
```

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+    HTML by default now. If you need the unescaped version, you can pass `escape: false`.
+
+    ```crystal
+    highlight("<p>This is a beautiful morning, but also a beautiful day</p>", "beautiful", escape: false)
+    ```
+
+    We've also fixed a few issues related to cookies.
+    * You can now set a `samesite` default.
+    * Exceptions aren't raised when trying to delete a cookie that doesn't exist
+    * Cookies are written at the correct time
```
paulcsmith: I'd leave this out just because I think it is super hard to explain what's going on and without the explanation it doesn't provide much value. What do you think?

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    ```crystal
+    class Posts::Index < BrowserAction
+      # Matches GET /posts
+      # GET /posts/2020
+      # GET /posts/2020/07
+      get "/posts/?:year/?:month" do
+        # ...
+      end
+    end
```
paulcsmith: Does this actually work? I think I added it to the router, but I don't think there is support in Lucky for extracting those parts and making a nilable param. So I *think* this should be removed until we add support to Lucky::Routable

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+    out thanks to [@matthewmcgarvey](https://github.com/luckyframework/lucky_router/pull/26)!
+
+    ## CLI Task Args
+
+    Writing CLI tasks for your application are pretty common, but they can also be a bit complicated.
+    Lucky already comes with a ton of tasks like `lucky db.migrate`, and `lucky gen.page About`. When
+    you start to write your own, you may need to pass in arguments from the CLI. We break these in to 3 different type.
+
+    * `arg` - This is the most standard CLI argument type. e.g. `-m User` or `--model=User`.
+    * `switch` - Just like `arg`, but with out a value. Returns `true` if this flag is passed. e.g. `-v`, `-h`
```
paulcsmith:
```suggestion
    * `switch` - Just like `arg`, but without a value. Returns `true` if this flag is passed. e.g. `-v`, `-h`
```

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+    * `positional_arg` - When you want to pass args that are not flags. Lucky uses this with generators like
```
paulcsmith, line 135: Great explanations of these options!

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+    ```
+
+    On top of all the neat routing features, we were also able to squeeze a little more performance
+    out thanks to [@matthewmcgarvey](https://github.com/luckyframework/lucky_router/pull/26)!
+
+    ## CLI Task Args
+
+    Writing CLI tasks for your application are pretty common, but they can also be a bit complicated.
+    Lucky already comes with a ton of tasks like `lucky db.migrate`, and `lucky gen.page About`. When
+    you start to write your own, you may need to pass in arguments from the CLI. We break these in to 3 different type.
```
paulcsmith: What do you think about making the association between these sentences a bit clearer somehow. Maybe something like this? 

> Lucky has many built-in tasks like `lucky db.migrate` and `lucky gen.page`, but often times an application needs their own custom tasks. Now Lucky makes it easier than ever to write these types of tasks

That doesn't quite work, but maybe sparks some ideas? Or maybe we cut it out and skip straight to something like: "Lucky now makes it easier than ever to write custom CLI tasks with arguments" and assume people will know what that means?

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+    * `switch` - Just like `arg`, but with out a value. Returns `true` if this flag is passed. e.g. `-v`, `-h`
+    * `positional_arg` - When you want to pass args that are not flags. Lucky uses this with generators like
+      `lucky gen.page About` where `About` is a positional arg.
+
+    ```crystal
+    # lucky search.reindex User
+    class Search::Reindex < LuckyCli::Task
+      summary "Reindex records for a model"
+    
+      positional_arg :model, "Specify which model to reindex", required: true
```
paulcsmith, line 143: I thought we changed this to be require by default? Maybe I'm misremembering, but might be worth checking

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+    # lucky generate_sitemaps --environment=dev
+    class GenerateSitemaps < LuckyCli::Task
+      summary "Generate some sitemaps"
+      arg :environment, "Specify the ENV for the sitemaps", format: /(dev|prod)/
+    
+      def call
+        # ...
+       end
+    end
+    ```
```
paulcsmith, line 179: Fantastic examples!!

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    ## Lucky Flow
+
+    The underlying shard that LuckyFlow wraps has been replaced with the selenium shard by
+    [@matthewmcgarvey](https://github.com/matthewmcgarvey/selenium.cr). This update gives us
+    a ton of flexibility by allowing us to support all browsers and not just Chrome. On top of that,
+    we can now stay up-to-date with the new W3C WebDriver standard going forward.
```
paulcsmith:
```suggestion
    [@matthewmcgarvey](https://github.com/matthewmcgarvey/selenium.cr). This update allows us to support all browsers and not just Chrome. On top of that,
    we can now stay up-to-date with the new W3C WebDriver standard going forward.
    
    It also automatically installs the correct driver for your version of Chrome so browser testing is easier than ever to get started with.
```

src/posts/lucky_0.23.cr
```diff
@@ -0,0 +1,219 @@
...
+
+    For the most part, everything still works as it did before, just with a shiny new engine!
+
+    ## Parting words
+
+    This release has seen a huge surge in community contributions. It's tough maintaining
+    open source projects, so we just want to say how much we appreciate all of the hard work
+    the community has put in to Lucky making this framework so amazing!
+
+    We're really excited about getting closer to a 1.0.0 release, and we can't do it without
```
paulcsmith:
```suggestion
    We're really excited about getting closer to a 1.0 release, and we can't do it without
```

paulcsmith (commented):
Thanks for writing this up. These blog posts are always tricky and a lot of work so I really appreciate it!

src/posts/lucky_0.23.cr
jwoertink: Good catch! We did, I just never updated my example I was looking at to remember how this worked 😂

src/posts/lucky_0.23.cr
jwoertink: oh... I didn't even check. I assumed it did 😅  yeah, I'll take that out.

paulcsmith (approved):
Looks good to go :D

PR #341, opened 2020-06-29
Jaw/v0.23.0

jwoertink:
This PR updates the actual website to use Lucky 0.23 and Crystal 0.35


jwoertink:
> Unsupported shard.lock. It was likely generated from a newer version of Shards.

I'm not sure what that means 🤔  maybe it's because Gh Actions is caching shards?

paulcsmith:
I think the shard.lock is broken because you updated the Dockerfile, but we're now using GH actions. So I think the GH action file needs to be updated and Dockerfile removed. I think that'll fix it!

shard.yml
```diff
@@ -8,27 +8,29 @@ targets:
...
     github: icyleaf/markd
     branch: master
   cadmium_transliterator:
     github: cadmiumcr/transliterator
     branch: master
+
+development_dependencies:
+  lucky_flow:
+    github: luckyframework/lucky_flow
+    version: ~> 0.7.0
```
paulcsmith: Could you add a newline here?

src/pages/home/index_page.cr
```diff
@@ -53,7 +53,7 @@ class Home::IndexPage
   private def render_feature_grid
     div class: "mx-auto container" do
       div class: "text-center lg:mx-5 md:text-left px-6 py-5 bg-white md:bg-transparent border-b border-grey-light md:border-none" do
-        mount Shared::FlashMessages.new(@context.flash)
+        m Shared::FlashMessages, @context.flash
```
paulcsmith, line 56: I like this new `m` syntax :D

paulcsmith (approved):
Thanks for doing this! LGTM!

PR #350, opened 2020-07-04
Better docs on using params

### LINKED ISSUE #249: Document the new param escape hatches
https://github.com/luckyframework/lucky/pull/1053

### LINKED ISSUE #312: Document that params.get 'trims' params. Document 'get_raw' for untrimmed values
https://github.com/luckyframework/lucky/pull/1144

jwoertink:
Fixes #312
Fixes #249

### DIFF
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -336,9 +336,16 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
 
     > Learn more about [error handling](#{Guides::HttpAndRouting::ErrorHandling.path}).
 
-    ## Query parameters
+    ## Handling parameters
 
-    When query parameters are passed to an action, you have access to these from the `params` method.
+    Parameters, or `params`, are data that is sent from client back to the server. There are a few different ways this can happen:
+
+    * Path parameters - The dynamic values passed in your route path. e.g. `/users/:id`.
+    * Query parameters - These are added to the end of a URL after the `?` in key/value pairs. e.g. `?page=1`
+    * Form parameters - When you perform an HTTP POST, the data sent through the body. This may be formatted as JSON, or form data
+    * Multipart parameters - Similar to form parameters, but generally to contain a file
+
+    When parameters are passed to an action, you have access to these with the `params` method.
 
     ```crystal
     # src/actions/users/index.cr
@@ -370,6 +377,56 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
 
     > By default, all param values are trimmed of blankspace. If you need the raw value,
     > use `params.get_raw(:key)` or `params.get_raw?(:key)`.
+
+    ### Params from query string
+
+    Returns just the query params as `HTTP::Params`.
+
+    You can use the `params.get(:key)` method to get your values from the query string,
+    or `params.get_raw(:key).
+
+    Lucky also gives you access to a `params.from_query["key"]` method which does the same thing.
+
+    ```crystal
+    # example.com?q=Lucky
+    params.get(:q) #=> "Lucky"
+    params.get_query["q"] #=> "Lucky"
+
+    params.get?(:search) #=> nil
+    params.get_query["search"]? #=> nil
+    ```
+
+    ### Params from JSON
+
+    Parses the request body as `JSON::Any` or raises `Lucky::ParamParsingError` if JSON is invalid.
+
+    ```crystal
+    # {"users": [{"name": "Skyler"}]}
+    params.from_json["users"][0]["name"].as_s #=> "Skyler"
+    ```
+
+    ### Params from form data
+
+    Returns x-www-form-urlencoded body params as `HTTP::Params`.
+
+    Generally you'd use `params.get()` for these, but for consistency sake, Lucky gives you this method.
+
+    ```crystal
+    params.get(:name)
+    params.from_form_data["name"]
+    ```
+
+    ### Params from multipart
+
+    Returns multipart params and files.
+
+    ```crystal
+    form_params = params.from_multipart.last # Hash(String, String)
+    form_params["name"]                      # "Kyle"
+
+    files = params.from_multipart.last # Hash(String, Lucky::UploadedFile)
+    files["avatar"]                    # Lucky::UploadedFile
+    ```
 
     ### Accepting specific params
 
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -420,7 +420,7 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     files["avatar"]                    # Lucky::UploadedFile
     ```
 
-    ### Accepting specific params
+    ### Type-safe query params
 
     Other times you may want to accept parameters in the query string, e.g. `/users?page=2`.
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -302,14 +302,49 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    ### Accepting specific params
+
     Other times you may want to accept parameters in the query string, e.g. `https://example.com?page=2`.
 
     ```crystal
     # src/actions/users/index.cr
     class Users::Index < BrowserAction
       param page : Int32 = 1
 
-      route do
```
jwoertink, line 348: Going to slowly start phasing these out.

paulcsmith: Good idea 👍

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -302,14 +302,49 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
 
     ## Query parameters
 
+    When query parameters are passed to an action, you have access to these from the `params` method.
```
paulcsmith: I'm not sure this belongs here. You can use `params.get` for any kind of param not just query params. Maybe a new section like `Dynamic param access` or something?

jwoertink: Yeah, we probably need a whole section to talk about the `from_json`, `from_query`, etc... I'll need to think about what the looks like.

paulcsmith: Oh yeah I forgot about those 😂 So many new things!

jwoertink: Ok, I'm going to refactor this section a little to add some more info.

jwoertink:
@paulcsmith ok, I tore this up a bit. I'm not too happy with the wording, but I don't really understand the usage for these param escape hatches. Just pushing this up for now so you can see the general direction, and we can figure out how to best make this make sense.

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -336,7 +336,99 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    class Users::Index < BrowserAction
+
+      # example.com/users?page=1&filter=active
+      get "/users" do
+        # These are both type `String`
+        filter = params.get(:filter)
+        page = params.get(:page)
+
+        # This is missing and will raise an error
+        per = params.get(:per)
```
edwardloveall: I think this is fine, but it made me want to try out an alternate format.
```suggestion
        filter = params.get(:filter) # type String
        page = params.get(:page) # type String
        per = params.get(:per) # Error! there is no parameter :per
```

You can pick which one you like best 😄

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -336,7 +336,99 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    * Form parameters - When you perform an HTTP POST, the data sent through the body. This may be formatted as JSON, or form data
+    * Multipart parameters - Similar to form parameters, but generally to contain a file
+
+    When parameters are passed to an action, you have access to these with the `params` method.
+
+    ```crystal
+    # src/actions/users/index.cr
+    class Users::Index < BrowserAction
+
+      # example.com/users?page=1&filter=active
```
edwardloveall: Something about this url being inside the example file feels odd to me. How about putting it up in a sentence before the code sample:

> Here's an example of using parameters when visiting the `/users?page=1&filter=active` path:

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -336,7 +336,99 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    ```
+
+    > By default, all param values are trimmed of blankspace. If you need the raw value,
+    > use `params.get_raw(:key)` or `params.get_raw?(:key)`.
+
+    ### Params from query string
+
+    Returns just the query params as `HTTP::Params`.
+
+    You can use the `params.get(:key)` method to get your values from the query string,
```
edwardloveall: It feels odd that this restates what you just did above. Maybe you could say something like "As seen above..." or maybe it doesn't matter because no one reads docs like a book 😆

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -336,7 +336,99 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    or `params.get_raw(:key).
+
+    Lucky also gives you access to a `params.from_query["key"]` method which does the same thing.
```
edwardloveall: This is also weird to me. Why would I use this if it does the same thing?

edwardloveall (approved):

src/actions/guides/http_and_routing/routing_and_params.cr
jwoertink: Yeah, my thought was if someone on the home page searches "params", then lands on that section, they might not read above it. It is a bit redundant, but maybe it's ok since most people jump around in docs?

src/actions/guides/http_and_routing/routing_and_params.cr
jwoertink: 🤷‍♂️  no clue 😅  My guess is just for consistency? So you're not doing `params.from_json` and `params.get` in the same method or something?

jwoertink:
Alright, I'm feeling a lot better about these docs. There's a ton more info here now.

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -336,16 +336,100 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+
+    ```crystal
+    form_params = params.from_multipart.last # Hash(String, String)
+    form_params["name"]                      # "Kyle"
+
+    files = params.from_multipart.last # Hash(String, Lucky::UploadedFile)
+    files["avatar"]                    # Lucky::UploadedFile
+    ```
+
+    ### Accepting specific params
```
paulcsmith: Thoughts on this?

```suggestion
    ### Type-safe query params
```

paulcsmith (approved):
Looks good. Just one tiny suggestion! LGTM when you are ready

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -336,16 +336,100 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    ```crystal
+    get "/users" do
+      per = params.get?(:per) # returns nil
+
+      plain_text "..."
+    end
+    ```
+
+    > By default, all param values are trimmed of blankspace. If you need the raw value,
+    > use `params.get_raw(:key)` or `params.get_raw?(:key)`.
```
paulcsmith: Actually just one thought. It might be good to add a not somewhere here that it is preferred to use the type-safe macro `param` and link to it below, so people realize they can use that instead and it is preferred 99% of the time

jwoertink: Ok, I think I see what you're saying. I'll need to think about how to word that.

jwoertink: Alright, I moved that section to the top,and reworded a few things. I think this flows much better now. Thanks!

jwoertink:
Thanks for the reviews team!

PR #355, opened 2020-07-06
Integrate HTML2Lucky converter

### LINKED ISSUE #76: Integrate HTML -> Lucky
I'm using this a ton lately. Now that this site is built in Lucky, we can integrate the two projects.

ref: https://github.com/luckyframework/html2lucky/issues/11#issuecomment-508175401

jwoertink:
Fixes #76 

This PR adds in all of the functionality from the https://github.com/luckyframework/html2lucky project. 

We gain a few benefits from this move:

* Allows us to match our existing design
* 1 less Lucky app to keep upgraded
* The current converter site takes up to a minute to load sometimes due to the Heroku plan. The lucky website is usually instant.
* More exposure to this tool will help ease people moving existing sites over
* Easier to manage issues and get PRs on this tool

### NOTE
Ok, so.... I'm not great at design, and tailiwind is super confusing to me. Here's what this tool looks like as of right now:
Screen Shot 2020-07-06 at 3 35 40 PM

Obviously we want this to look better, but design is also not a show stopper. That's an easy thing to fix afterward too. But any ideas for some quick design cleanup are welcome.


src/utils/html2lucky/tags/tag.cr
```diff
@@ -0,0 +1,82 @@
+abstract class HTML2Lucky::Tag
```
edwardloveall: What do you think about putting these in a separate directory? If this were rails I'd say `lib` but that's obviously taken. Maybe `/src/lib`? `/vendor`? They just feel odd to me being in the main app code, even under `utils`. What do you think?

Thinking about it more, maybe it makes sense for this to become a shard, just for the separation of concerns, not because we expect anyone to ever use it like a shard. Out of scope for this PR though.

src/pages/converter/index_page.cr
```diff
@@ -0,0 +1,30 @@
+class ConvertHtmlToLucky::IndexPage < PageLayout
+  needs input : String
+  needs output : String
+
+  def content
+    div class: "my-4" do
+      h2 @title, class: "text-3xl mb-4"
+      para "Lucky uses Crystal classes and methods to generate HTML. It may sound crazy at first, but the advantages are numerous. Never accidentally print nil to the page, extract and share partials using regular methods. Easily read an entire page by looking at just the render method. Text is automatically escaped for security. And it’s all type safe. That means no more unmatched closing tags, and never rendering a page with missing data."
+
+      div class: "flex my-4" do
```
edwardloveall: This is untested but I think this will give the boxes a little bit of room from each other. The other problem is that they aren't the same height. Try this:

```suggestion
      div class: "flex my-4 space-x-4 items-stretch" do
```

edwardloveall (approved):
Thanks for bringing this over!

src/utils/html2lucky/tags/tag.cr
jwoertink: Well, in rails, the `lib` directory is meant for libs that your project needs, but there isn't a direct gem for. Maybe Lucky needs a directory like that in  `src/`. I know a few of my projects have a `src/lib` for that same reason. Would that be too confusing though?

src/pages/converter/index_page.cr
jwoertink: yup. Gives a little space between them

src/utils/html2lucky/tags/tag.cr
edwardloveall: Yeah I worry about that too, which is why I offered `/vendor` as an alternative. Might be worth trying, and then eventually shipping a separate shard.

paulcsmith (approved):
Looks great! I can take a look at the design later :D

jwoertink:
Awesome! Thanks for the reviews team!

PR #356, opened 2020-07-08
Fixed typos and data type

guglielmogcg:
In response pagination pages headers section


jwoertink (approved):
Ah, good catch with the `to_s` calls. Thanks for submitting this!

PR #365, opened 2020-07-10
Move to Lexbor HTML parser

### LINKED ISSUE #358: It only converts tags inside the <body>
HTML to convert
```html
<!DOCTYPE html>
<html lang="en">

<head>

  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
  <meta name="description" content="">
  <meta name="author" content="">

  <title>Theme</title>

  <!-- Custom fonts for this theme -->
  <link href="vendor/fontawesome-free/css/all.min.css" rel="stylesheet" type="text/css">
  <link href="https://fonts.googleapis.com/css?family=Montserrat:400,700" rel="stylesheet" type="text/css">
  <link href="https://fonts.googleapis.com/css?family=Lato:400,700,400italic,700italic" rel="stylesheet" type="text/css">

  <!-- Theme CSS -->
  <link href="css/theme.min.css" rel="stylesheet">

</head>

<body id="page-top">
  <h1>Hi</h1>
</body>
</html>
```

Lucky response
```crystal
h1 "Hi"
```

jwoertink:
The lexbor parser has a smaller footprint, and parsers faster, but has all of the exact same limitations as myhtml. One extra benefit we get from this PR is that we can now parse `<head>` tags. Before this PR, if you tried to convert `<meta content="">`, it wouldn't return anything. With this PR it will now return `meta content: ""`. 

This PR was originally intended to close #358 but turns out that won't be possible. See note in the issue.

TL;DR - we would need to hand roll a custom parser to parse full documents.

Instead, I've just added a note mentioning the limitations. You can parse markup in the `<head>` tags, or in the `<body>` tag, but not both at the same time.


jwoertink:
> /bin/sh: 1: cmake: not found

I give up 🤷‍♂️ Not sure if it's worth the trouble of adding in cmake and all that if this isn't going to add much benefit. I can probably get myhtml to do the same thing, but I'm pretty burned out on this at this point.

PR #366, opened 2020-07-13
Change to shorter /html URL

paulcsmith:
I think it'd be nice to have a shorter URL that people can easily remember and type in.

This also make the namespace a REST resource and changes the folder name to match it


src/actions/html_conversions/new.cr
```diff
@@ -0,0 +1,8 @@
+class HtmlConversions::New < BrowserAction
```
paulcsmith, line 1: Renamed from `Index` to `New`

jwoertink (approved):
Sweet 👍  I like it.

Issue #211, opened 2020-02-11
Document new options for `lucky gen.model`


paulcsmith:
https://github.com/luckyframework/lucky/pull/1009

jwoertink:
This is done.

PR #367, opened 2020-07-15
Add documentation section around generating models with columns

stephendolan:
There's actually no way to know that you can provide column names to `lucky gen.model` in the website documentation, and you'd need to run `lucky gen.model --help` to figure it out!

Coming from Rails, I thought it'd be nice in that initial "Generating a Model" section to at least show something that would pop up in search results.

Not sure if y'all would prefer the positioning further up in the section or not, I kind of thought of this as a footnote to the section.


jwoertink:
I think this fixes https://github.com/luckyframework/website/issues/211

paulcsmith (approved):
Looks great! Thank you

paulcsmith:
Closes #211

PR #374, opened 2020-07-21
Updating the guides on memoize.

### LINKED ISSUE #332: Update memoization docs
https://luckyframework.org/guides/http-and-routing/routing-and-params#memoization

https://github.com/luckyframework/lucky/pull/1139

jwoertink:
Fixes #332

### DIFF
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -306,10 +306,29 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     end
     ```
 
-    There are a few caveats to the `memoize` macro. It will not memoize `false`, or `nil` return values.
-    If your method returns a "falsey" value, then it will be ran each time. Another thing to note is
-    you can't memoize a method that takes an argument. This is due to the dynamic nature of the
-    arguments.
+    You can also memoize methods that have a `nil` or `false` return value like when finding a record.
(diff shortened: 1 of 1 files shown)


src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -306,10 +306,29 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     end
     ```
 
-    There are a few caveats to the `memoize` macro. It will not memoize `false`, or `nil` return values.
-    If your method returns a "falsey" value, then it will be ran each time. Another thing to note is
-    you can't memoize a method that takes an argument. This is due to the dynamic nature of the
-    arguments.
+    You can also memoize methods that have a `nil` or `false` return value like when finding a record.
+    As well as the ability to pass arguments to your memoized method.
```
paulcsmith: The second sentence seems like a continuation of the first but I think it is a separate sentence. Maybe something like:

```suggestion
    Memoized methods can return any value, including`nil` or `false`.
    You can also pass arguments to your memoized method.
```

paulcsmith (approved):
Just one suggestion. LGTM!

src/actions/guides/http_and_routing/routing_and_params.cr
jwoertink: Yeah, I wasn't quite sure how to word that. I think your suggestion is good.

PR #376, opened 2020-07-23
Adding a bit more to the route generation docs. 

### LINKED ISSUE #254: Better docs on using route helper methods
We're not really clear on how you would pass query params using route helper methods. 

https://luckyframework.org/guides/http-and-routing/link-generation
https://luckyframework.org/guides/http-and-routing/routing-and-params#query-parameters

Neither of these pages really show you that you would do `Users::Index.with(page: 1)`. They also don't really tell you that you can do `Guides::Database.path(anchor: "middle-of-page")`.

jwoertink:
Fixes #254

This cleans up the routing helper guides to show a few more examples to be a little more clear on how this stuff works. Also added a link to this guide from the params guide since many people will land there first.

### DIFF
--- a/src/actions/guides/http_and_routing/link_generation.cr
+++ b/src/actions/guides/http_and_routing/link_generation.cr
@@ -14,7 +14,7 @@ class Guides::HttpAndRouting::LinkGeneration < GuideAction
     You can access them as class methods on the action itself. They are:
 
     * `route` - returns a `Lucky::RouteHelper` object that contains the path, HTTP method,
-      , and the full URL as methods `path`, `method`, `url`.
+      and the full URL as methods `path`, `method`, `url`.
     * `with` - an alias for `route` that is used for passing parameters.
     * `path` - returns a string of just the path.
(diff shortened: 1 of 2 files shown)


src/actions/guides/http_and_routing/link_generation.cr
```diff
@@ -9,38 +9,77 @@ class Guides::HttpAndRouting::LinkGeneration < GuideAction
...
+        to: Guides::Show.with(id: "making-tacos", anchor: "cheeses")
+    end
+    ```
+
+    ### Using raw strings
 
-    * `Projects::Users::Index.path(project_id: "my_project_id")` and it will return ->
-      `"/projects/my_project_id/users"`
-    * `Projects::Users::Index.with(project_id: "my_project_id")` and it will return a
-      `LuckyWeb::RouteHelper` whose `#path` method returns
```
jwoertink, line 38: `LuckyWeb` 😂

paulcsmith (approved):
Looks great!

PR #380, opened 2020-07-28
Updating heroku docs to mention running tasks.

### LINKED ISSUE #257: Update heroku deploy guide to mention seeding required seeds
Since we come with a task to seed your DB with required seeds, we should also document how to do this.

Probably just after the Uploading the app section.

https://luckyframework.org/guides/deploying/heroku

jwoertink:
Fixes #257

Also added a note about the release phase that makes migrations run automagically.


paulcsmith (approved):
Looks great!

PR #384, opened 2020-08-06
add info about using `Int32` for `avram_enum` in database table

lxxxvi:
Hey there 👋 

I couldn't find any information about the database data type for `avram_enum`, would it make sense to add it here?


jwoertink (approved):
Hey! Thanks for submitting this, and welcome to Lucky! 

This is a great point. It's good to be clear that we need to add an Int32 column for it to work. 👍

PR #385, opened 2020-08-07
Add missing select_count documentation

stephendolan:
Adding back what seems to be a missing header, and inserting in some missing documentation surrounding `select_count`.


jwoertink (approved):
I'm surprised we missed this 😅  Thanks for adding it in!

PR #387, opened 2020-08-08
Replace old "render" calls with "html"

### LINKED ISSUE #1233: Fallback section still mentions render
The [fallback section](https://luckyframework.org/guides/http-and-routing/routing-and-params#fallback-routing) of the website still mentions `render Home::IndexPage` when it should be `html Home::IndexPage`

stephendolan:
Closes https://github.com/luckyframework/lucky/issues/1233!

Searched for all occurrences of render using regex: `rg "render\s[A-Z]{1}"`, and this looks like the last two!


jwoertink (approved):
👍

PR #391, opened 2020-08-08
Add documentation for using "autogenerated: true" with timestamp columns

### LINKED ISSUE #232: Document `autogenerated: true` for created_at/updated_at
https://github.com/luckyframework/avram/issues/315#issuecomment-596135588

stephendolan:
This PR is intended to close #232, which spawned from https://github.com/luckyframework/avram/issues/315

Hopefully with this quick callout and some copy-ready code to use for custom behavior, we can avoid future confusion around this.


src/actions/guides/database/models.cr
```diff
@@ -118,6 +118,15 @@ class Guides::Database::Models < GuideAction
     end
     ```
 
+    Note that should you remove the `timestamps` default column and manually add `created_at` and/or `updated_at` to specific models, you will need to leverage the `autogenerated: true` flag in order to preserve the automatic default behavior. An example can be seen in the default `timestamps` macro implementation in Avram:
```
jwoertink: This sentence confused me. I may have to think about this a bit 🤔

stephendolan: Lol reading this again, I agree. I'll take another swing at the language this week unless y'all can think of a succinct way to put it.

jwoertink: *summoning* @edwardloveall 😬

edwardloveall: Can I talk though this a little? I think what you're trying to say could be phrased like:

> If you remove `timestamps` from these default columns and you still want the automatic default behavior, you'll need to ensure that:
>
> * Add `created_at` and/or `updated_at` columns to specific models
> * Make sure they are annotated with `autogenerated: true`
>
> Here's an example from Avram:

---

I'm not totally sure what that does or if I got the instructions right. Seems like the problem here is mostly that it's a run-on sentence with some conditions at the start _and_ the end. Breaking it up helps.

jwoertink: Yeah, I think the whole purpose of the `autogenerated: true` is that it makes it so the SaveOperation doesn't yell when you don't pass those fields. If a column is required, but the user will never pass the value to a SaveOperation, then we add the `autogenerated: true`, and internally it all handles itself. (same with the primary key I think).

edwardloveall: Ah neat. That all makes sense. Might be worth adding a bit in there about that:

> If these fields are not marked as `autogenerated: true` operations will complain about these required fields being missing. The operation can't know that the database will generate these values automatically, so we `autogenerated: true` to give it that information.

stephendolan:
We did it!

Screen Shot 2020-08-10 at 6 23 06 PM

jwoertink (approved):
Oh yeah, that looks a lot better 👍  Great work!

src/actions/guides/database/models.cr
```diff
@@ -118,6 +118,20 @@ class Guides::Database::Models < GuideAction
     end
     ```
 
+    If you remove timestamps from these default columns and you still want the automatic default behavior, you'll need to:
```
paulcsmith:
```suggestion
    If you remove `timestamps` from `default_columns` and you still want the automatic default behavior, you'll need to:
```

src/actions/guides/database/models.cr
```diff
@@ -118,6 +118,20 @@ class Guides::Database::Models < GuideAction
...
+
+    - Add created_at and/or updated_at columns to specific models
```
paulcsmith:
```suggestion
    - Add `created_at` and/or `updated_at` columns to specific models
```

src/actions/guides/database/models.cr
```diff
@@ -118,6 +118,20 @@ class Guides::Database::Models < GuideAction
...
+    - Make sure they are annotated with `autogenerated: true`
```
paulcsmith:
```suggestion
    - Set the `autogenerated` option to `true`
```

paulcsmith (commented):
A couple small suggestions. Overall looks great!

stephendolan:
Sorry for all of the code fencing omissions, @paulcsmith! This should be good to go. I appreciate everyone's time on this.

paulcsmith:
No prob! Thanks for working on this!

PR #393, opened 2020-08-08
Remove "route" and "nested_route" code samples

### LINKED ISSUE #279: Remove route inference from the guides and use explicit routes
Part of: https://github.com/luckyframework/lucky/discussions/789#discussioncomment-3296

And also change to not use the parent resource as a module namepsace in routes: https://github.com/luckyframework/lucky/discussions/789#discussioncomment-2855

`Users::Index` not `Projects::Users::Index`

stephendolan:
This PR aims to close #279 

In the future, it is planned to remove `route` and `nested_route` helpers for routes. The issue associated with this PR links to GitHub Discussions that dive into much greater detail.

To facilitate that, this PR does two things:

- Refactors all calls to `route` and `nested_route` to use explicit routing
- Refactors the HTTP and Routing section to not reference or note the availability of `route` or `nested_route` with the goal of pushing new users to adopt explicit routing as their default

**Edit**

This PR is being refocused on only removing the example code using `route` and `nested_route`, while still leaving the actual documentation around those methods intact.

### DIFF
--- a/src/actions/guides/database/validating-saving.cr
+++ b/src/actions/guides/database/validating-saving.cr
@@ -164,7 +164,7 @@ class Guides::Database::ValidatingSaving < GuideAction
 
     ```crystal
     class Users::Create < BrowserAction
-      route do
+      post "/users" do
         # params will have the form params sent from the HTML form
         SaveUser.create(params) do |operation, user|
           if user # if the user was saved
@@ -429,8 +429,8 @@ class Guides::Database::ValidatingSaving < GuideAction
 
     ```crystal
     class Posts::Comments::Create < BrowserAction
-      route do
-        post = PostQuery.find(id)
+      post "/posts/:post_id/comments" do
+        post = PostQuery.find(post_id)
         # Params contain the title and body, but not the post_id
         # So we set it ourselves
         SaveComment.create(params, post_id: post.id) do |operation, comment|
@@ -602,7 +602,7 @@ class Guides::Database::ValidatingSaving < GuideAction
 
     ```crystal
     class Searches::Create < BrowserAction
-      route do
+      post "/searches" do
         SearchData.new(params).submit do |operation, results|
           # `valid?` is defined on `operation` for you!
           if operation.valid?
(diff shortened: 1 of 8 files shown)


src/actions/guides/database/validating-saving.cr
```diff
@@ -428,9 +428,9 @@ class Guides::Database::ValidatingSaving < GuideAction
     In this example, we'll associate a comment with a post:
 
     ```crystal
-    class Posts::Comments::Create < BrowserAction
```
jwoertink, line 431: This is probably fine to leave since it's namespaced in the route.

jwoertink (commented):
I think this is a great move. Though, now that I'm looking at it, I'm wondering if we might confuse people from the generated apps still using `route` and `nested_route`. Maybe we hold off on merging? Let's see what @paulcsmith says

paulcsmith (commented):
Looks great! But I agree with @jwoertink that we should hold off until we remove it from Lucky itself. Right now people still need the docs since `route` is fairly common.

Maybe to ship this earlier this could be split into 2 separate PRs

1. Change all examples to use `get/put/post` like you've done here. But just change the examples and leave the docs/examples on `route/nested_route`. This way people get used to using `get/put/post/etc.` but still can get docs on `route/nested_route`
2. Separate PR that moves `route/nested_route` to the very bottom of the guides and mentions that it is deprecated but possible to use with a shard. This way people can still get the docs, and can use `route` if they want to. This PR would have to come later once we've actually removed the `route/nested_route` macros

Up to you though. We can also leave this as-is and merge later

stephendolan:
I like the idea of not letting this stagnate and getting something merged in to move the needle in the right direction! I'll resubmit with the changes to src/actions/guides/http_and_routing/routing_and_params.cr reverted!

stephendolan:
This is ready to merge, @paulcsmith / @jwoertink.

I'll create an issue momentarily to follow up with the secondary action of removing (or just moving) the actual documentation for `route` and `nested_route`.

paulcsmith (approved):
Lovely! This makes me really like the decision to remove `route/nested_route`. So much easier to tell what the route is and what params are available now!

PR #394, opened 2020-08-09
Add target "_blank" to external navigation links

### LINKED ISSUE #204: Link to docs
Now we have docs! https://luckyframework.github.io/lucky/

We should link to them and also get docs generated for lucky_flow, avram, carbon, and habitat so we can link to those as well. We should also figure out if we can tag gh-pages so we can link to official releases of docs and not just master

stephendolan:
This PR adds a "target" attribute for navbar links, so that any links not pointing to luckyframework.org will go to a new tab. It felt kind of easy to be navigated away from the site, which could lead to folks not exploring as thoroughly as they normally would have.


jwoertink (approved):
That makes sense to me! 👍

paulcsmith:
I'd like to hold off on this for a bit. The nav bar is fairly dense and I'd like to test it out and make sure it still works at various sizes.

Also I think I'd like to take a little different approach. My thought is that we change `Guides` to `Learn`. Clicking learn takes you to a page with links to the guides, links to API docs for all the libs (Avram, Lucky, etc. There are a lot!) and eventually screencasts

Something like this offset grid in TailwindUI (but without icons)

Screen Shot 2020-08-10 at 3 19 05 PM


If you have access to TailwindUI and want to give this a go that'd be great! If not I'll try to tackle this sometime soon(ish). I think to start it could be just sections for the guides and then sections for all the API docs

stephendolan:
I just recently acquired TailwindUI, @paulcsmith, so I'm happy to take a swing at it!

This was mostly to get the ball rolling since I think it adds a huge benefit to new users to be able to explore the API docs that they may not know exist, otherwise.

paulcsmith:
Right on! TailwindUI rocks :) LMK if you run into anything or need to bounce ideas off me. Happy to help

stephendolan:
@paulcsmith / @jwoertink,

This will be a much larger effort, so I've reverted this to just contain the `target` additions, which I still think are helpful but aren't necessarily related to the new "Learn" section I'm prototyping.

I'll also probably collect feedback via Discord / GH Discussions before actually opening a PR to ensure it doesn't go stale.

jwoertink:
Ok, I think that's fine. Thanks for updating!

PR #402, opened 2020-08-13
Document optional path parameters

### LINKED ISSUE #339: Document optional path param routing
https://luckyframework.org/guides/http-and-routing/routing-and-params

https://github.com/luckyframework/lucky_router/pull/18

stephendolan:
This PR aims to close #339

This section ended up being pretty minimal, but it exposes the API and (I think) provides a nice example where you may want to use it, since the purpose escaped me initially.

### DIFF
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -117,6 +117,26 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     `project_id` and `task_id` method generated on the class for accessing the named
     parameters.
 
+
+    ### Optional path parameters
+
+    Sometimes it can be helpful to allow optional parameters in a route's path. We can accomplish this by prefixing a path parameter with a `?`, like this:
+
+    ```crystal
+    # src/actions/posts/index.cr
+    class Posts::Index < BrowserAction
+      get "/posts/:year/:month/?:day" do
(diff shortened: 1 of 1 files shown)


paulcsmith (approved):
This is an excellent example. Thanks!

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -117,6 +117,26 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+        if day
+          plain_text "I'll show all posts on a specific day!"
+        else
+          plain_text "I'll show all posts in a given month!"
+        end
+      end
+    end
+    ```
+
+    In the above example, we require that the `Post` index route has both a `:year` and `:day` provided, but allow users to optionally route to a specific `:day` as well.
```
stephendolan, line 138: Darn it, @paulcsmith, I just realized that this should be "`:year` and `:month`" 😭 

Should I open a PR to correct or is there a way to change that first `:day` to `:month` in master from your end?

paulcsmith: I can change it pretty easily on master :D

stephendolan: Thanks man, sorry about that! I think it was off just enough that my brain filtered it out 😛

PR #405, opened 2020-08-14
Add HTMLConverter access through public API

### LINKED ISSUE #404: Add API interface for luckyframework.org/html
Allow for non-authenticated requests to call the `html2lucky` method directly from an API exposed on the website. This could be used for neat community tooling like what I'd like to build for VSCode here:
https://github.com/stephendolan/vscode-lucky/issues/4

stephendolan:
This PR aims to close #404 with an API that returns 200s, not 404s as the issue number may suggest.

A few key components:
- Added a good old-fashioned `accepted_formats[:json]` to `ApiAction` to bring that file up to date with default generated Lucky apps, so that we don't have to specify allowed formats for all API actions in the future.
- Added a new `POST` route at `/api/html_conversions` that takes in a string called `input`, and returns a JSON payload with one top-level key called `content`, containing the HTML `input` converted to crystal. This mimics the structure and functionality already established in the browser action at `/src/actions/html_conversions/{create,new}.cr`.

Because our `ApiAction` already didn't include `Api::Auth::RequireAuthToken`, there was no need to add a specific authentication skipping include.

### DIFF
--- a/src/actions/api/html_conversions/create.cr
+++ b/src/actions/api/html_conversions/create.cr
@@ -1,0 +1,19 @@
+class Api::HtmlConversions::Create < ApiAction
+  before halt_if_html_conversion_disabled
+
+  param input : String = ""
+
+  post "/api/html_conversions" do
+    output = HTML2Lucky::Converter.new(input).convert
+
+    json({content: output}, :ok)
+  end
+
+  private def halt_if_html_conversion_disabled
+    if ENV["DISABLE_HTML_CONVERSION_API"]?
+      head :service_unavailable
+    else
+      continue
+    end
+  end
+end
--- a/spec/requests/api/html_conversions/create_spec.cr
+++ b/spec/requests/api/html_conversions/create_spec.cr
@@ -1,0 +1,17 @@
+require "../../../spec_helper"
+
+describe Api::HtmlConversions::Create do
+  it "returns a 200 response with converted content" do
+    response = AppClient.exec(Api::HtmlConversions::Create.with(input: "<div>Test</div>"))
+
+    response.should send_json(200, content: "div \"Test\"\n")
+  end
+
+  it "returns a 503 (Service Unavailable) when HTML conversions are disabled" do
+    ENV["DISABLE_HTML_CONVERSION_API"] = "Shut it down!"
+
+    response = AppClient.exec(Api::HtmlConversions::Create.with(input: "<div>Test</div>"))
+
+    response.status_code.should eq(503)
+  end
+end
--- a/spec/setup/start_app_server.cr
+++ b/spec/setup/start_app_server.cr
@@ -1,0 +1,9 @@
+app_server = AppServer.new
+
+spawn do
+  app_server.listen
+end
+
+Spec.after_suite do
+  app_server.close
+end
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -11,5 +11,6 @@ require "./support/**"
 require "./setup/**"
 
 include Carbon::Expectations
+include Lucky::RequestExpectations
 
 Habitat.raise_if_missing_settings!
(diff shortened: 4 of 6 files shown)


paulcsmith (commented):
Nice! I think it'd be good to add a simple test https://luckyframework.org/guides/testing/testing-actions to make sure it works as expected and returns things in the correct format.

Also, I think it'd be nice to wrap the action in an ENV var and so we can disable it if we do get too much traffic. Probably with a `before :halt_if_converted_disabled` that returns `head :service_unavailable` if the ENV var is set. I'm guessing something like `DISABLE_HTML_CONVERTER_API` or something

src/actions/api/html_conversions/create.cr
```diff
@@ -0,0 +1,9 @@
+class Api::HtmlConversions::Create < ApiAction
+  param input : String = ""
+
+  post "/api/html_conversions" do
+    output = HTML2Lucky::Converter.new(input).convert
+
+    json({content: output}, HTTP::Status::OK)
```
paulcsmith: I'm not sure, but I think you can do this:

```suggestion
    json({content: output}, :ok)
```

stephendolan: Neat!

stephendolan:
Thoughts on something like this, @paulcsmith? The `before` action felt a bit easy to miss, but we can go that route if it's a better convention:

```crystal
class Api::HtmlConversions::Create < ApiAction
  param input : String = ""

  post "/api/html_conversions" do
    output = HTML2Lucky::Converter.new(input).convert

    if html_conversion_api_disabled
      head :service_unavailable
    else
      json({content: output}, :ok)
    end
  end

  private def html_conversion_api_disabled
    ENV["DISABLE_HTML_CONVERSION_API"]?
  end
end
```

paulcsmith:
@stephendolan I like the `before` because it is outside of the normal flow. Kind of an "filter" of sorts. The other thing is that with the proposed approach it would still be converting the HTML. Of course you can change the conditional, but it kind of shows how it can be easy to accidentally do processing that you didn't expect to do. Using a `before` makes sure _nothing_ after it will run and take up server resources.

stephendolan:
Ah, right, this would kind of defeat the purpose of even having the toggle... and agreed, that miss highlights the value in `before` quite well!

stephendolan:
@paulcsmith The specs will fail for this since I can't quite seem to figure out how to test that environment variable toggle... pushing to this branch in case you're able to provide some insight or assistance. It's at least got all of the suggested refactors and missing tests scaffolded out!

paulcsmith:
Looks like specs are failing. Not 100% sure why but my guess is that the config/server.cr is missing some config for starting the app. Either that or something is missing in spec helper for starting the app. You can look at the lucky CLI repo to check for missing stuff. I’m in my phone so can’t look now but if you get stuck lmk and I’ll give it a look later

stephendolan:
What a journey! The tests took twice as long a the code, but we've forged a path ahead for those that come behind us.

I had to add a `spec/setup/start_app_server.cr`, and everything (even the `503` sad path test) now passes! Huzzah!

Thanks for the tip, Paul, that would have taken a lot longer to track down without it.

paulcsmith (approved):
Looks great! Thanks so much. And yes I agree this will make it much easier for people moving forward!

paulcsmith:
Also I’ll be deploying this with the next release of lucky so won’t quite be ready for a bit

stephendolan:
Thanks for the heads up, @paulcsmith, that's totally fine. Sorry you had to be so involved with this PR - my excitement for a new VSCode feature got the better of me!

paulcsmith:
No problem at all! I’m very excited about the integration!

stephendolan:
@paulcsmith / @jwoertink Do y'all have an ETA on when this'll be available in prod?

I'd love to get https://github.com/stephendolan/vscode-lucky/pull/5 merged and deployed now that v0.24.0 is out 😄

jwoertink:
Yeah, I'd say probably around [object Object] or so. Thank you for using JBot 4.0.

PR #406, opened 2020-08-15
Password less dbms

mwlang:
Documents how to use password-less database connections for Avham

### DIFF
--- a/src/actions/guides/database/database_setup.cr
+++ b/src/actions/guides/database/database_setup.cr
@@ -27,6 +27,9 @@ class Guides::Database::DatabaseSetup < GuideAction
     * port
 
     > You can also set a `DATABASE_URL` environment variable. (e.g. `postgres://root@localhost:5432/my_database`)
+    [components](#{Guides::Frontend::RenderingHtml.path}).
+
+    > If you wish to use password-less connections for local development, please see [Installing Postgres](#{Guides::GettingStarted::Installing.path(anchor: Guides::GettingStarted::Installing::ANCHOR_POSTGRESQL)})
 
     ### Connection Pool
 
--- a/src/actions/guides/database/database_setup.cr
+++ b/src/actions/guides/database/database_setup.cr
@@ -27,7 +27,6 @@ class Guides::Database::DatabaseSetup < GuideAction
     * port
 
     > You can also set a `DATABASE_URL` environment variable. (e.g. `postgres://root@localhost:5432/my_database`)
-    [components](#{Guides::Frontend::RenderingHtml.path}).
 
     > If you wish to use password-less connections for local development, please see [Installing Postgres](#{Guides::GettingStarted::Installing.path(anchor: Guides::GettingStarted::Installing::ANCHOR_POSTGRESQL)})
 
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -225,6 +225,25 @@ class Guides::GettingStarted::Installing < GuideAction
 
     There are other installation methods available in [Postgres CLI tools docs](https://postgresapp.com/documentation/cli-tools.html)
 
+    ### 1b. Password-less logins for local development
+
+    If you wish to use PostgreSQL w/o a password, you'll need to ensure your user is added to the `pg_hba.conf` file.  We recommend
+    adding this entry right after the `postgres` user entry:
+
+    ```plain
+    local   all             postgres                                peer
+
+    # TYPE  DATABASE        USER            ADDRESS                 METHOD
+    host    all             your_username   127.0.0.1/32            trust  # Add this line
+    ```
+
+    > `your_username` is the local user you're logging into the DBMS with.  Change as appropriate.
+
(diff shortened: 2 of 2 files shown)


src/actions/guides/database/database_setup.cr
```diff
@@ -27,6 +27,9 @@ class Guides::Database::DatabaseSetup < GuideAction
     * port
 
     > You can also set a `DATABASE_URL` environment variable. (e.g. `postgres://root@localhost:5432/my_database`)
+    [components](#{Guides::Frontend::RenderingHtml.path}).
```
jwoertink: Is this misplaced?

src/actions/guides/getting-started/installing.cr
```diff
@@ -225,6 +226,25 @@ class Guides::GettingStarted::Installing < GuideAction
 
     There are other installation methods available in [Postgres CLI tools docs](https://postgresapp.com/documentation/cli-tools.html)
 
+    ### 1b. Password-less logins for local development
```
jwoertink: I've never had to do this on Mac, and I always use password-less DBs.... Should this specify Linux?

jwoertink (commented):
This is good info, but I'm worried about it being a lot of non-lucky related info. I'm afraid newcomers (jr devs and such) may see all this extra setup stuff and end up in a worse position, or get turned off to using Lucky from trying to set something up. 

What OS are you running where you had to set this up?

paulcsmith (commented):
I'm thinking maybe this could go in a "Troubleshooting" section at the bottom, and then we could add a note with a link to it in the Postgres section. "Having trouble setting up passwordless Postgres? Check out [link to troubleshooting anchor]

Thoughts?

BTW, this is super helpfuL! Thank you!

src/actions/guides/database/database_setup.cr
mwlang: ugh! bad editor!  I'm trying out VS Code and it's a little weird how it does keyboard-driven selecting.

src/actions/guides/getting-started/installing.cr
mwlang: Yes, we probably should call out Linux and I will make that change -- I wasn't 100% sure if it applied to new macOS setups or not.

mwlang (commented):
As long as the user gets some sort of indication in the right places in their set up experience, I'm happy.

src/actions/guides/database/database_setup.cr
mwlang: this is removed now.

src/actions/guides/getting-started/installing.cr
mwlang: Singling out Linux now.  I double-checked Homebrew formula for PostgreSQL and it is setting up the `pg_hba.conf` file to be a lot more lenient, `trust`ing all logged in users to connect to the server.  Homebrew's choice for defaults doesn't make a lot of sense for Linux since macOS is desktop 99% of the time while Linux is more likely to be a server for PostgreSQL deploys.  Hence the difference in user experience.

mwlang: One thing I would add.  I am a long-time macOS user as well as Linux on servers.  Linux as a desktop/development experience is fairly new for me and I would never deploy PostgreSQL without strong passwords on all the users.  That's what completely threw me here because I was used to macOS' defaults as well as simply operating in server-space the way a security conscious devop should.   Because of all the changes with the Crystal language, I suspected the issue resided in implementation rather than system configuration for many hours, before looking into postgresql itself.  That's the pain I'm hoping to solve for the next Linux-based developer with this PR.

src/actions/guides/getting-started/installing.cr
```diff
@@ -1,5 +1,6 @@
 class Guides::GettingStarted::Installing < GuideAction
   ANCHOR_INSTALL_REQUIRED_DEPENDENCIES = "perma-install-required-dependencies"
+  ANCHOR_POSTGRESQL                    = "perma-install-postgres"
```
paulcsmith, line 3: For this to work you'll also need to use `#{permalink(ANCHOR_POSTGRESQL)` above the heading on 211. See example on line 11

src/actions/guides/getting-started/installing.cr
```diff
@@ -225,6 +226,26 @@ class Guides::GettingStarted::Installing < GuideAction
 
     There are other installation methods available in [Postgres CLI tools docs](https://postgresapp.com/documentation/cli-tools.html)
 
+    ### 1b. (Linux only) Password-less logins for local development
+
+    Homebrew installed PostgreSQL on macOS are configured by default to allow password-less logins.  But for Linux, if you wish to
+    use PostgreSQL w/o a password, you'll need to ensure your user is added to the `pg_hba.conf` file with `trust` method specified.
```
paulcsmith: The `w/o` may not be familiar to non-native English speakers so probably better to go with `without` just in case 

```suggestion
    use PostgreSQL without a password, you'll need to ensure your user is added to the `pg_hba.conf` file with `trust` method specified.
```

paulcsmith (commented):
2 tiny things then LGTM!

paulcsmith:
Looks great! Thank you :D

PR #407, opened 2020-08-15
Update links for Discord

edwardloveall:
I didn't bother changing old blog posts because I feel like those probably should remain an artifact in time.


stephendolan:
👀

PR #416, opened 2020-09-06
Update crystal and lucky versions in the guides

jwoertink:
Updates the versions shown in the guides


src/actions/guides/getting-started/installing.cr
```diff
@@ -89,7 +89,7 @@ class Guides::GettingStarted::Installing < GuideAction
     * Install Crystal with `asdf`.
 
     ```plain
-    asdf install crystal latest
+    asdf install crystal #{LuckyCliVersion.compatible_crystal_version}
```
jwoertink, line 92: Better to show the exact version supported instead of latest just in case.

PR #419, opened 2020-09-09
Update guides with Avram Credentials

### LINKED ISSUE #382: Update guides with Avram Credentials
Also document all the things you can do with it

https://github.com/luckyframework/avram/pull/433

jwoertink:
Fixes #382

### DIFF
--- a/src/actions/guides/database/database_setup.cr
+++ b/src/actions/guides/database/database_setup.cr
@@ -15,20 +15,61 @@ class Guides::Database::DatabaseSetup < GuideAction
 
     > If you're using the postgres app on macOS, be sure to install [CLI tools](https://postgresapp.com/documentation/cli-tools.html).
 
+    ### Avram Configuration
+
+    Avram comes with an `Avram::Credentials` class for configuring your database credentials. This helps to catch incorrect settings at compile-time
+    like malformatted usernames and such.
+
+    To configure Lucky to connect to your database, open up your `config/database.cr` file. You'll find a few options within the `AppDatabase` configure block.
+
     ### Standard Options
 
-    To configure Lucky to connect to your database, open up your `config/database.cr` file.
-    You'll find a few standard options within the `AppDatabase` configure block.
+    Set the `settings.credentials` to `Avram::Credentials.new` and pass in these options:
 
-    * database
-    * hostname
-    * username
-    * password
-    * port
+    * database : `String` - This is the name of your database. The default is set at the top of this file.
+    * hostname : `String?` - The host where your database is located. Generally "localhost".
+    * username : `String?` - Your database user.
+    * password : `String?` - Your password.
+    * port : `Int32?` - The port to connect to. Default is `5432`
+    * query : `String?` - A query string of connection pool settings.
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/database_setup.cr
```diff
@@ -15,20 +15,61 @@ class Guides::Database::DatabaseSetup < GuideAction
...
+    ```
+
+    On most systems, you can leave the password blank if your setup doesn't require a password. If you wish to use password-less connections for local development,
+    and leaving the password blank doesn't work, please see [Installing Postgres](#{Guides::GettingStarted::Installing.path(anchor: Guides::GettingStarted::Installing::ANCHOR_POSTGRESQL)})
+    for more tips.
+
+    ### Using Postgres connection string
+
+    Avram also supports using a standard connection string whether you want to use unix sockets or connect with credentials. For this, we can just set
+    an environment variable like `DATABASE_URL` and parse it with `Avram::Credentials.parse`
```
paulcsmith: Might be worth mentioning that this is already in config/database.cr. The way it is phrased now makes it seem like they would need to do this themselves.

src/actions/guides/database/database_setup.cr
```diff
@@ -41,18 +82,38 @@ class Guides::Database::DatabaseSetup < GuideAction
     * retry_attempts
     * retry_delay
 
-    To set the connection pool options, just append a query string to the end of the
-    `settings.url` option (e.g. `?initial_pool_size=5&retry_attempts=2`)
+    To set the connection pool options, just set the `query` option in your `Avram::Credentials` to a query string.
+    (e.g. `initial_pool_size=5&max_pool_size=10`).
```
paulcsmith:
```suggestion
    (e.g. `query: "initial_pool_size=5&max_pool_size=10"`).
```

src/actions/guides/database/database_setup.cr
```diff
@@ -41,18 +82,38 @@ class Guides::Database::DatabaseSetup < GuideAction
...
 
-    ### Avram Configuration
+    > If using an environment variable, the query string should come at the end of your connection string.
+    > (e.g. `postg)
```
paulcsmith: I'm not sure I understand this example. Maybe a full database url example would help?

src/actions/guides/database/database_setup.cr
```diff
@@ -41,18 +82,38 @@ class Guides::Database::DatabaseSetup < GuideAction
...
+    end
+    ```
 
     Optionally, the `lazy_load_enabled` is set to `false` for development and test.
     This causes Lucky to raise an exception if you forget to preload an association,
     but will not raise an exception in production.
 
+    ```crystal
+    settings.lazy_load_enabled = Lucky::Env.production?
+    ```
```
paulcsmith: Sounds like this should be under a new header since it is not part of  "Apps not using Avram"

src/actions/guides/database/database_setup.cr
jwoertink: 😂 I must have been in the middle of writing it and got side tracked. Oops!

paulcsmith (approved):

PR #421, opened 2020-09-11
Adding docs for the new data response method. 

### LINKED ISSUE #372: Document 'data' method
https://github.com/luckyframework/lucky/pull/1220

jwoertink:
Fixes #372

### DIFF
--- a/src/actions/guides/http_and_routing/request_and_response.cr
+++ b/src/actions/guides/http_and_routing/request_and_response.cr
@@ -164,7 +164,7 @@ class Guides::HttpAndRouting::RequestAndResponse < GuideAction
     ### Rendering raw data
 
     If you need to return a file that already exists, you can use `file`, but in the case that you only have the
-    data to be written to a file, you can use this `data` method.
+    String data, you can use this `data` method.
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/http_and_routing/request_and_response.cr
```diff
@@ -160,6 +161,26 @@ class Guides::HttpAndRouting::RequestAndResponse < GuideAction
     end
     ```
 
+    ### Rendering raw data
+
+    If you need to return a file that already exists, you can use `file`, but in the case that you only have the
+    data to be written to a file, you can use this `data` method.
```
paulcsmith:
```suggestion
    String data, you can use this `data` method.
```

I think this was a bit misleading since it won't be written to a file on the server, it'll just send the data to the client which may or may not save it to a file

jwoertink: yeah, this was a little confusing. I had troubling figuring out what this even did.. This reads better though. Thanks!

paulcsmith (approved):

PR #425, opened 2020-09-21
optimized version of github actions CI

skojin:
optimized version of github actions CI
* cache is used for node
* shards are cached

in current version cache is used only for ~/.cache/crystal

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,23 +28,30 @@ jobs:
         node-version: '12.x'
     - name: "Install yarn"
       run: npm install -g yarn
+
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn cache dir)"
+    - name: Set up Yarn cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+        key: ${{ runner.os }}-yarn--${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-yarn--
+
+    - name: Set up Crystal cache
+      uses: actions/cache@v2
+      with:
+        path: |
+          lib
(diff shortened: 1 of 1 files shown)


script/setup
```diff
@@ -20,7 +20,7 @@ yarn dev | indent
 print_done
 
 notice "Installing shards"
-shards install --ignore-crystal-version | indent
+shards check --ignore-crystal-version || shards install --ignore-crystal-version | indent
```
jwoertink: I'm not sure about this one. What is the `check` being used for? One thing I've noticed with `shards check` is if you change the `shard.yml` like removing a shard no longer used, the `shards check` command still passes.

.github/workflows/ci.yml
```diff
@@ -28,23 +28,31 @@ jobs:
...
     - name: "Install yarn"
       run: npm install -g yarn
-    - name: Setup Lucky
-      run: script/setup
-    - name: Cache Crystal
-      uses: actions/cache@v1
+
+    - name: Get yarn cache directory path
+      id: yarn-cache-dir-path
+      run: echo "::set-output name=dir::$(yarn cache dir)"
```
jwoertink, line 34: When setting up the Yarn cache, should it just set the `id` the same as the Crystal cache does?

jwoertink (commented):
Thanks for catching this. Just a few comments to clarify some changes here.

script/setup
skojin: shards install always fetch shards, I used check to prevent this. But I not checked it for remove shards, it says "Dependencies are satisfied" when remove shards and change shard version, so better remove this check

.github/workflows/ci.yml
skojin: actually id: crystal-cache not need here too, because it not used
it used in config like this https://github.com/skojin/lucky-bug-honeypot/blob/master/.github/workflows/ci.yml#L87
in if command

jwoertink (approved):
This looks great! Thanks for adding this in!

PR #426, opened 2020-09-22
Add missing `sudo` in Ubuntu documentation

vjousse:
A `sudo` is missing to be able to `chown deploy:deploy` the app directory.

### DIFF
--- a/src/actions/guides/deploying/ubuntu.cr
+++ b/src/actions/guides/deploying/ubuntu.cr
@@ -55,7 +55,7 @@ class Guides::Deploying::Ubuntu < GuideAction
 
     ```bash
     sudo mkdir /srv/<yourapp>
-    chown deploy:deploy /srv/<yourapp>
+    sudo chown deploy:deploy /srv/<yourapp>
     ```
 
     ## Create a PostgreSQL database


src/actions/guides/deploying/ubuntu.cr
```diff
@@ -55,7 +55,7 @@ class Guides::Deploying::Ubuntu < GuideAction
 
     ```bash
     sudo mkdir /srv/<yourapp>
-    chown deploy:deploy /srv/<yourapp>
+    sudo chown deploy:deploy /srv/<yourapp>
```
jwoertink, line 58: does this also need to be `sudo chown -R deploy:deploy /srv/<yourapp>`? Or does it not matter if it's recursive?

vjousse: No it doesn't have to be recursive as the directory is supposed to be empty: it was created with the `mkdir` just above.

jwoertink: oh right 😅  still too early here. I wasn't finished with ☕  yet 😛

jwoertink (approved):

PR #428, opened 2020-09-22
Be consistent with DATABASE_URL value

vjousse:
Above in the documentation, we are using `postgresql://<username>:<password>@127.0.0.1/<appname>_production` for the `DATABASE_URL` value. I think we should be consistent here too as it's the format that will work for most people.

### DIFF
--- a/src/actions/guides/deploying/ubuntu.cr
+++ b/src/actions/guides/deploying/ubuntu.cr
@@ -167,7 +167,7 @@ class Guides::Deploying::Ubuntu < GuideAction
     User=deploy
     Environment="LUCKY_ENV=production"
     Environment="SECRET_KEY_BASE=<random unique key>"
-    Environment="DATABASE_URL=postgres:///<yourapp>_production"
+    Environment="DATABASE_URL=postgresql://<username>:<password>@127.0.0.1/<appname>_production"
     Environment="HOST=127.0.0.1"
     Environment="PORT=5000"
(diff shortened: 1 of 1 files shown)


src/actions/guides/deploying/ubuntu.cr
```diff
@@ -167,7 +167,7 @@ class Guides::Deploying::Ubuntu < GuideAction
     User=deploy
     Environment="LUCKY_ENV=production"
     Environment="SECRET_KEY_BASE=<random unique key>"
-    Environment="DATABASE_URL=postgres:///<yourapp>_production"
+    Environment="DATABASE_URL=postgresql://<username>:<password>@127.0.0.1/<appname>_production"
```
jwoertink: Interesting. I've always done `postgres://`, but I just read 

> The URI scheme designator can be either postgresql:// or postgres://.

https://www.postgresql.org/docs/12/libpq-connect.html

I'm thinking maybe we should just keep it consistent with what Avram uses [here](https://github.com/luckyframework/avram/blob/master/src/avram/credentials.cr#L132) 🤔  Can you grep through the rest of the docs and see if we mix this up? If we use both, then lets roll with the shorter one, but I like the rest of the change here 👍

vjousse: It's `postgres` that is used everywhere in the docs. I've updated the PR accordingly :+1:

paulcsmith (approved):
This looks great. Thanks @vjousse

PR #429, opened 2020-09-23
Escape text that contains a quote when converting HTML to Lucky

vjousse:
When you submit HTML that contains quotes inside the text, the lucky code generated is not correct.
```html
<div>Hello "world"</div>
```
produces:
```crystal
div "Hello "world""
```

it should produce:
```crystal
div "Hello \"world\""
```

This PR fixes this problem


jwoertink (approved):
Sweet! This is what I love about having the converter in this repo now. So much easier to fix bugs with it 😂

PR #435, opened 2020-10-01
Fix Markdown Regex Formatting Issue

### LINKED ISSUE #434: Weird syntax highlighting issue
If you look at the code example on this page https://luckyframework.org/guides/command-line-tasks/custom-tasks#positional-args you'll see a weird syntax highlighting related to the Regex.

stephendolan:
This PR aims to resolve #434

The issue is likely related to the markdown lib since the last `+` in `/\\w+:\\w+/`.

While we should probably open an issue there, I was able to resolve this and fix a side-effect of the same code by adding line begin/end anchors (`^` and `$`).

This prevents me from using this task as `lucky gen.model MyModel some_column:anotherl33th4ck3r`, which would pass for the non-achored regex.


jwoertink (approved):
works for me if you've seen it fixed locally!

PR #437, opened 2020-10-01
Add detailed Carbon Adapter documentation

### LINKED ISSUE #432: Add documentation for Carbon::DevAdapter's `print_email` argument
It's possible to print the content of your emails sent through `DevAdapter` straight to your logs for investigation.

In fact, I submitted a PR to make this the default here:
https://github.com/luckyframework/lucky_cli/pull/555

As noted by Paul in that PR, this probably won't be necessary once Breeze and Pulsar are handling things for us. That said, the email guides are quite scant with a "See Carbon README" right now and this might be a good chance to add some helpful tips and tricks for debugging and testing.

stephendolan:
This PR aims to close #432.

It breaks out a separate "Adapters" section, explaining each currently-supported adapter in greater detail, as well as specific options for the DevAdapter like `print_email`.

Here's a capture of the new page layout/content:

new_adapter_documentation

### DIFF
--- a/src/actions/guides/emails/sending-emails-with-carbon.cr
+++ b/src/actions/guides/emails/sending-emails-with-carbon.cr
@@ -7,10 +7,9 @@ class Guides::Emails::SendingEmailsWithCarbon < GuideAction
 
   def markdown : String
     <<-MD
-    ## Carbon Configuration File
+    ## Configuring Email
 
-    Lucky generates a default email configuration file in `config/email.cr`. In that file you
-    can add SendGrid keys and change adapters.
+    Lucky leverages the [Carbon](https://github.com/luckyframework/carbon) library for writing, sending, and testing emails. Carbon can be configured using the default file generated with a new Lucky application in `config/email.cr`. In that file you can add SendGrid keys and change adapters.
 
     ## Adapters
     
@@ -23,6 +22,7 @@ class Guides::Emails::SendingEmailsWithCarbon < GuideAction
     There are two ways to leverage the `DevAdapter`. The first is by telling the adapter to simply capture all Carbon output without printing or displaying the email content, which is the default:
 
     ```crystal
+    # config/email.cr
     BaseEmail.configure do |settings|
       settings.adapter = Carbon::DevAdapter.new
     end
@@ -31,6 +31,7 @@ class Guides::Emails::SendingEmailsWithCarbon < GuideAction
     If you want to see your email content printed to your development or test server logs, you can use the optional `print_emails` flag:
 
     ```crystal
+    # config/email.cr
     BaseEmail.configure do |settings|
       settings.adapter = Carbon::DevAdapter.new(print_emails: true)
     end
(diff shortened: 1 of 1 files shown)


src/actions/guides/emails/sending-emails-with-carbon.cr
```diff
@@ -7,11 +7,53 @@ class Guides::Emails::SendingEmailsWithCarbon < GuideAction
...
+
+    ```crystal
+    BaseEmail.configure do |settings|
+      settings.adapter = Carbon::SendGridAdapter.new(api_key: ENV["SEND_GRID_KEY"])
+    end
+    ```
+
+    ### AWS SES Adapter
+
+    The `AwsSesAdapter` by community member [keizo3](https://github.com/keizo3/carbon_aws_ses_adapter), once configured with your AWS credentials, will send all emails through the [AWS SES](https://aws.amazon.com/ses/) email service.
```
jwoertink: I like the idea; however, this shard hasn't been updated in a long time, and my guess is, due to the restricted crystal version, it wouldn't "bundle" (shard? 🤔 ). We should probably add this to that "cookbook" page though so we can at least highlight that something exists. 

What would be cool though is maybe a section that explains how to create your own adapter, but that could come in a later PR.

src/actions/guides/emails/sending-emails-with-carbon.cr
```diff
@@ -7,11 +7,53 @@ class Guides::Emails::SendingEmailsWithCarbon < GuideAction
...
+    Carbon supports a growing number of adapters thanks to contributions from the community.
+
+    ### Dev Adapter
+    
+    The `DevAdapter` ships with Carbon by default, and is useful for handling emails in a development or test environment. It can also be leveraged in production to effectively disable emails.
+
+    There are two ways to leverage the `DevAdapter`. The first is by telling the adapter to simply capture all Carbon output without printing or displaying the email content, which is the default:
+
+    ```crystal
+    BaseEmail.configure do |settings|
```
jwoertink, line 26:
```suggestion
    # config/email.cr
    BaseEmail.configure do |settings|
```

This just gives a small clarity on where someone is making this change at.

src/actions/guides/emails/sending-emails-with-carbon.cr
```diff
@@ -7,11 +7,53 @@ class Guides::Emails::SendingEmailsWithCarbon < GuideAction
 
   def markdown : String
     <<-MD
-    ## Configuring Carbon
+    ## Carbon Configuration File
```
jwoertink: Since these titles are indexed in to the search algolia, I wonder if this should actually say something like "Configuring Emails", and then have a section that explains what "Carbon" is. What are your thoughts? I would be good either way.

jwoertink (commented):
Nice! I dig the direction, and this page definitely needed the love. Just a few comments then it'll be good to go 👍

src/actions/guides/emails/sending-emails-with-carbon.cr
stephendolan: That's a great idea, Jeremy! I'll push a change and see how it feels to read through.

src/actions/guides/emails/sending-emails-with-carbon.cr
```diff
@@ -7,10 +7,48 @@ class Guides::Emails::SendingEmailsWithCarbon < GuideAction
 
   def markdown : String
     <<-MD
-    ## Configuring Carbon
+    ## Configuring Email
 
-    Lucky generates default configuration in `config/email.cr`. In that file you
-    can add SendGrid keys and change adapters.
+    Lucky leverages the [Carbon](https://github.com/luckyframework/carbon) library for writing, sending, and testing emails. Carbon can be configured using the default file generated with a new Lucky application in `config/email.cr`. In that file you can add SendGrid keys and change adapters.
```
jwoertink, line 12: Yup! I like it 👍

jwoertink (approved):

PR #448, opened 2020-10-21
Fix small typo in installing.cr

heyfixit:
Caught a small typo while reading docs, one of those where you rephrase a sentence but accidentally leave a piece of the old one.


matthewmcgarvey (approved):
thanks @heyfixit!

PR #485, opened 2020-11-09
Renames app_client to api_client

ianmaxwell:
* noticed that app_client has been recently renamed to api_client in https://github.com/luckyframework/lucky_cli/pull/534


jwoertink (approved):
Ah good catch! I was supposed to open an issue about this and must have forgot. Thanks for fixing it 👍

PR #505, opened 2020-12-07
Allow rendering raw HTML with Cmark

### LINKED ISSUE #504: Enum permalink is wrong
The link to the Enum section [in the Column-Types](https://luckyframework.org/guides/database/models#column-types) (find on page "see using enum") section points to a permalink "#perma-using-enum", but the actual section id is just "#using-enums"

Similarly, https://luckyframework.org/guides/database/validating-saving#perma-saving-enums and https://luckyframework.org/guides/database/querying-records#perma-querying-enums both from https://luckyframework.org/guides/database/models#using-enums (find on page "To learn more about using enums")

stephendolan:
Closes #504.

When we switched to CMark, we didn't realize that it would substitute raw HTML for a comment:
https://github.com/amauryt/cr-cmark-gfm/blob/master/src/cmark/renderers/html_renderer.cr#L182-L190

We leverage raw HTML to support permalinks on pages:
https://github.com/luckyframework/website/blob/master/src/actions/guide_action.cr#L22-L24

To get around that behavior, we pass the `Unsafe` option to the HTML renderer. Details in the API docs here:
https://amauryt.github.io/cr-cmark-gfm/Cmark/Option.html#Unsafe

I **did** verify that a `<script>` tag, for example, is still rendered within quotes and not executed, since it's in the list of filtered tags specified in the GFM spec:
https://github.github.com/gfm/#example-653

### DIFF
--- a/src/utils/custom_markdown_renderer.cr
+++ b/src/utils/custom_markdown_renderer.cr
@@ -28,7 +28,7 @@ class CustomMarkdownRenderer
   end
 
   def self.html(content)
-    options = Cmark::Option.flags(ValidateUTF8, Smart)
+    options = Cmark::Option.flags(ValidateUTF8, Smart, Unsafe)
     html = Cmark.gfm_to_html(content, options)
     HtmlAutolink.new(html).call
   end


jwoertink (approved):
🎉

PR #515, opened 2020-12-17
Blog post to promote the release of 0.25

jwoertink:
This is the blog post to announce the release of 0.25.0! 🎉 

It's a long one, so I had to cut it short at the end 😂

### DIFF
--- a/src/posts/lucky_0.25.cr
+++ b/src/posts/lucky_0.25.cr
@@ -263,14 +263,14 @@ class Lucky025Release < BasePost
     end
     ```
 
-    This means you can also now do a has_many though a has_many through 😬
+    Related to this change, we've also fixed bugs that limited the types of associations that could be used for the "through" association 😬
 
     ### Models now support SQL VIEW
 
-    Generally when we think of models, we think of database tables. The models even have a `table()` method to denote this!
+    Generally when we think of models, we think of database tables. The models even have a `table()` method to denote this.
     Well now there's a `view()` method as well!
 
-    > SQL VIEWs are like tables, but generally READ-ONLY, and have no primary key. Their data is usually aggregated from other table sources.
+    > SQL VIEWs are like tables, but generally READ-ONLY, and might not have a primary key. Their data is usually aggregated from other table sources.
 
     ```crystal
     class AdminUser < BaseModel
@@ -281,15 +281,16 @@ class Lucky025Release < BasePost
     end
     ```
 
-    All of the columns for your `view` must be manually defined. These models do not come with any sort of `primary_key`, or `id` methods. This also means that methods like
-    `AdminUserQuery.find()`, and `admin_user.delete` won't work since these all rely on there being an `id` method. You must implement any of these methods yourself that you may
-    need.
+    All of the columns for your `view` must be manually defined. These models do not come with any sort of `primary_key` or `timestamps` by default. If your view has these columns, you will need to explicitly add them.
+    View models will not have a `SaveOperation` defined as they are meant to be read-only, and if no primary key is added, some features will be missing from the model and `BaseQuery`.
+    For example, `AdminUserQuery.find()` and `admin_user.reload` won't work since these rely on an `id` method.
+    You must implement any of these methods yourself if you need them.
 
     ### Routing Changes
 
-    The LuckyRouter got a few new fancy that we're super stoked about! The first one is "glob" routing.
+    The LuckyRouter got some fancy upgrades that we're super stoked about! The first one is "glob" routing.
 
-    Glob routes are basically a route where you know the first part in the path, but the end of the route path is a variable length. For example:
+    Glob routes are a route where the first part in the path is known, but the end of the route path is a variable length. For example:
 
     ```
     myblog.com/posts
@@ -298,7 +299,7 @@ class Lucky025Release < BasePost
     myblog.com/posts/2020/12/25
     ```
 
-    These routes may all just display posts, and you don't want to duplicate routes or pages for the same actions being taken. In this case, we can define a glob route
+    If these routes all display posts, it's not useful to duplicate routes or pages over multiple actions. In this case, we can define a glob route
     to catch all of these in to the same action.
 
     ```crystal
@@ -311,11 +312,10 @@ class Lucky025Release < BasePost
     end
     ```
 
-    Another great update to the router is that we can now catch duplicate routes defined. This helps you to catch mistakes in development when you're working on a larger
-    code base, and accidentally defined two conflicting routes.
+    Another great update to the router is that Lucky can now catch overridden (duplicate) routes. This helps catch mistakes in development, especially in large project where hundreds or more routes are defined.
 
     ```crystal
-    # These will now raise an error
+    # These will now raise an error when starting the app
 
     class Api::Posts::Show < ApiAction
       get "/api/posts/:id" do
@@ -330,9 +330,9 @@ class Lucky025Release < BasePost
     end
     ```
 
-    ### So many more!
+    ### And much more!
 
-    This blog post could go on for days to include all the awesome stuff we've added. Here's a super quick run down of a few more:
+    This blog post could go on for days to include all the awesome stuff we've added. Here's a quick run down of a few more:
 
     * Updates and new features to LuckyFlow like hover over element
     * More bug fixes in associations and file handling
@@ -340,22 +340,22 @@ class Lucky025Release < BasePost
     * Testing CLI Tasks is easier
     * Lots of cleanup, and refactors
 
-    Read through the [CHANGELOG](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md) to see what else!
+    Read through the [CHANGELOG](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md) to see it all!
 
     ## Parting words
 
     This is another step towards 1.0, and we're super stoked. We can't do this without continued support from our community.
-    The more hands and eyes on the project, the more fine tuned we can make this project!
+    The more hands and eyes on the project, the more fine tuned Lucky can be!
 
     Please give it a spin and help us find bugs so our next release is even more solid.
-    If you find any issues, don't hesitate to [report the issue](https://github.com/luckyframework/lucky/issues).
+    If you find any issues, don't hesitate to [report them](https://github.com/luckyframework/lucky/issues).
     If you're unsure, just hop on [Discord chat](https://discord.gg/HeqJUcb) so we can help you out.
 
     Thanks so much for the support!
 
     ### Follow and spread the word
 
-    If you haven't already, give us a [star on github](https://github.com/luckyframework/lucky),
+    If you haven't already, give us a [star on GitHub](https://github.com/luckyframework/lucky),
     and be sure to follow us on [Twitter](https://twitter.com/luckyframework/).
 
     For questions, or just to chat, come say hi on [Discord](https://discord.gg/HeqJUcb).
--- a/src/posts/lucky_0.25.cr
+++ b/src/posts/lucky_0.25.cr
@@ -131,9 +131,6 @@ class Lucky025Release < BasePost
     end
     ```
 
-    Lastly, there were some instances where you may be updating a record, and expecting a callback when everything was done. However, in the case of
-    a record that doesn't need to update anything, you'd never have any callbacks that fired. (i.e. kick off background job, or email, etc...)
-
     We've added one additional callback called `after_completed`. This callback is always called when the operation is successful (the internal `save_status` attribute is set to `:saved`).
     This means that even if the record never touches the database, the `after_completed` callback will still run.
 
--- a/src/posts/lucky_0.25.cr
+++ b/src/posts/lucky_0.25.cr
@@ -1,0 +1,370 @@
+class Lucky025Release < BasePost
+  title "Lucky v0.25 is hot off the press, and there's a lot!"
+  slug "lucky-0_25-release"
+
+  def published_at : Time
+    Time.utc(year: 2020, month: 12, day: 19)
+  end
+
+  def summary : String
+    <<-TEXT
+    Lucky v0.25 is out. A ton has changed,
+    and we're full steam ahead to 1.0!
+    TEXT
+  end
+
+  def content : String
+    <<-MD
+    [Lucky v0.25](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md#changes-in-025)
+    is out now, and we have a ton to go over. Let's get to it!
+
+    ### How to upgrade
+
+    See the [UPGRADE_NOTES](https://github.com/luckyframework/lucky/blob/master/UPGRADE_NOTES.md#upgrading-from-024-to-025).
+
+    Or you can view the changes using [LuckyDiff](https://luckydiff.com/?from=0.24.0&to=0.25.0).
+    This tool helps see what has changed between versions and will be included in upgrade notes from now on.
+
+    ## Here's what's new
+
+    Since the last release we've welcomed two more developer to the core team, [Matthew](https://github.com/matthewmcgarvey) and [Stephen](https://github.com/stephendolan).
+    These two have both been huge contributors in the commuinity, and since bringing them on, the movement towards 1.0 has more than doubled in speed!
+
+    Over the course of the last few months, we've made a ton of changes for this release, so let's break down a couple of the more notible changes:
+
+    ### `Avram::Operation` got a facelift
+
+    When you need to handle some logic that may not be tied directly to a single model, you can create an `Avram::Operation` to handle this. You've probably seen
+    examples in your app like the `RequestPasswordReset` and `SignInUser` that come with an Auth app.
+
+    Prior to this update, it was sort of "wild west", and completely up to you on how to implement these. For consistency sake, we recommended defining a `submit`
+    method, and then returning `yield self, value`. These operations were also limited with no way to use callbacks, file attributes, or define errors not tied to
+    an attribute.
+
+    In this release, we've created a whole new interface!
+
+    ```crystal
+    class PromoteUserValidator < Avram::Operation
+      param_key :user
+      needs user : User
+      attribute token : String
+
+      # before callbacks
+      before_run do
+        if user.has_inactive_account?
+          # custom errors
+          add_error(:user_inactive, "My custom error message")
+        end
+
+        validate_required(token)
+      end
+
+      # after callbacks
+      after_run do |user|
+        NotifyUser.new(user.email).deliver
+      end
+
+      # unified interface
+      def run
+        if user.is_promotable?
+          user
+        else
+          nil
+        end
+      end
+    end
+
+    PromoteUserValidator.run(params) do |operation, user|
+      if user
+        # the user must be promotable
+      else
+        # no user, lets check our custom error
+        operation.errors[:user_inactive]
+      end
+    end
+    ```
+
+    ### Better callbacks in `Avram::SaveOperation`
+
+    We've also made quite a few changes to the `Avram::SaveOperation` family as well.
+
+    Prior to this update, your `after_save` and `after_commit` callbacks couldn't be used with an anonymous block like how `before_save` works. Well,
+    now they can!
+
+    ```crystal
+    class SaveUser < User::SaveOperation
+      before_save do
+        # before save
+      end
+
+      after_save do |saved_user|
+        # after save
+      end
+
+      after_commit do |saved_user|
+        # after DB commit
+      end
+    end
+    ```
+
+    We have also added the ability to do these callbacks conditionally. This allows you to specify if the callbacks should actually run or not.
+
+    ```crystal
+    class SaveUser < User::SaveOperation
+      before_save :validate_card_number, if: :new_transaction?
+
+      after_save :update_billing, unless: :card_still_valid?
+
+      private def validate_card_number
+      end
+
+      private def new_transaction?
+        true
+      end
+
+      private def update_billing
+      end
+
+      private def card_still_valid?
+        false
+      end
+    end
+    ```
+
+    Lastly, there were some instances where you may be updating a record, and expecting a callback when everything was done. However, in the case of
+    a record that doesn't need to update anything, you'd never have any callbacks that fired. (i.e. kick off background job, or email, etc...)
+
+    To help with this, we've added 1 additional callback called `after_completed`. This callback is always called when the `save_status` is set to `:saved`.
+    This means that even if the record never touches the database, the `after_completed` callback will still run.
+
+    ```crystal
+    class SaveUser < User::SaveOperation
+
+      after_commit do |updated_user|
+        # This is only called if something on the user actually changed
+        AfterCommitJob.perform(updated_user.id)
+      end
+
+      after_completed do |updated_user|
+        # This is always called if the user is saved
+        MetricsUpdaterJob.perform(updated_user.id)
+      end
+
+    end
+    ```
+
+    ### `with_defaults` is now `tag_defaults`
+
+    When you create components like `Shared::Field`, you can use the `tag_defaults` method to apply default attributes to a set of HTML elements.
+    This method was called `with_defaults` before, but has been renamed for some clarity on what it does.
+
+    ### Changes to the Query objects
+
+    The `Query` objects also got a few new changes. The main one being that queries no longer mutate the object. Prior to this release, appending any
+    query method would mutate the original object. This made building queries really nice and easy, but presented a problem that many faced. For example:
+
+    ```crystal
+    q = UserQuery.new
+    if current_user.admin?
+      q.admin(true)
+    end
+
+    # this would fail because we've already mutated the query
+    # on the count, and can no longer query with the username ordering
+    user_total_count = q.select_count
+    users = q.username.asc_order
+    ```
+
+    In that case, you had to make sure you cloned the query, and they were in the correct order. Now we no longer mutate the query, but this requires changing your
+    code just slightly.
+
+    ```crystal
+    q = UserQuery.new
+    if current_user.admin?
+      q = q.admin(true)
+    end
+
+    user_total_count = q.select_count
+    users = q.username.asc_order
+    ```
+
+    With this change, this also means that defining default queries in your query class `initialize` need to change. To set a default, you'll now use the `defaults` method.
+
+    ```crystal
+    class UserQuery < User::BaseQuery
+      def initialize
+        defaults &.admin(true)
+      end
+    end
+    ```
+
+    ### Type-safe WHERE "OR"
+
+    This was one of the most requested methods we got! We held off since adding this in makes SQL queries quite a bit more complex, but we finally got it! (**sort of... see note below)
+
+    ```crystal
+    # WHERE users.name = 'Billy' OR users.name = 'Kelly'
+    UserQuery.new.name("Billy").or(&.name("Kelly"))
+    ```
+
+    As you can see, you call the `or` method which passes in the instance of the query object allowing you to chain additional WHERE clauses.
+
+    **NOTE: The above example works great, and will cover plenty of the queries that you may have, but we currently make no assumptions on where to place parenthesis for scoping
+    order of operations. Take this for example:
+
+    ```crystal
+    UserQuery.new.name("Billy").or(&.name("Kelly").age.gte(51)).or(&.admin(true))
+    ```
+
+    This query would generate `WHERE name = 'Billy' OR name = 'Kelly' AND age >= 51 OR admin = true`, and that may not give you the result you expect. We are working on a solution,
+    but for now if you need more complex control, you can pass raw SQL to `UserQuery.new.where("name = ? OR (name = ? && age >= ?)", "Billy", "Kelly", 51)`.
+
+    ### Model Association updates
+
+    When you're using a RDBMS like postgres, table associations become very important for properly structuring your data. There were several bugs fixed when it comes to
+    associations giving us a lot more power and control over our data.
+
+    To start, we will look at the `belongs_to` association method. It's pretty common to name your method the same as the model it's referencing, but in some cases,
+    you may want your association method to be named slightly different like this:
+
+    ```crystal
+    class Employee < BaseModel
+      table do
+        belongs_to boss : Manager
+      end
+    end
+    ```
+
+    But this would throw an error due to how the query methods were generated. We would be expecting to generate a method named `where_managers`, but that doesn't exist here.
+
+    With this release, we no longer make the assumption on what you're naming your associations. This also means that a small convention must be followed where your `where_` query
+    methods will append the name of the association.
+
+    ```crystal
+    EmployeeQuery.new.where_boss(ManagerQuery.new)
+    ```
+
+    To go along with the `belongs_to`, we've also fixed some bugs related to using `has_many through`. This update will require a slightly different syntax.
+
+    ```crystal
+    # Before update
+    class User < BaseModel
+      table do
+        has_many posts : Post
+
+        # you specified the Symbol of the has_many method to query through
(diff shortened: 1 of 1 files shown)


src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    See the [UPGRADE_NOTES](https://github.com/luckyframework/lucky/blob/master/UPGRADE_NOTES.md#upgrading-from-024-to-025).
+
+    Or you can view the changes using [LuckyDiff](https://luckydiff.com/?from=0.24.0&to=0.25.0).
+    This tool helps see what has changed between versions and will be included in upgrade notes from now on.
+
+    ## Here's what's new
+
+    Since the last release we've welcomed two more developer to the core team, [Matthew](https://github.com/matthewmcgarvey) and [Stephen](https://github.com/stephendolan).
+    These two have both been huge contributors in the commuinity, and since bringing them on, the movement towards 1.0 has more than doubled in speed!
```
matthewmcgarvey:
```suggestion
    These two have both been huge contributors in the community, and since bringing them on, the movement towards 1.0 has more than doubled in speed!
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    Over the course of the last few months, we've made a ton of changes for this release, so let's break down a couple of the more notible changes:
+
+    ### `Avram::Operation` got a facelift
+
+    When you need to handle some logic that may not be tied directly to a single model, you can create an `Avram::Operation` to handle this. You've probably seen
+    examples in your app like the `RequestPasswordReset` and `SignInUser` that come with an Auth app.
```
matthewmcgarvey:
```suggestion
    examples in your app like the `RequestPasswordReset` and `SignInUser` that come with an app generated with authentication.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+      private def card_still_valid?
+        false
+      end
+    end
+    ```
+
+    Lastly, there were some instances where you may be updating a record, and expecting a callback when everything was done. However, in the case of
+    a record that doesn't need to update anything, you'd never have any callbacks that fired. (i.e. kick off background job, or email, etc...)
+
+    To help with this, we've added 1 additional callback called `after_completed`. This callback is always called when the `save_status` is set to `:saved`.
```
matthewmcgarvey: Seems like this comment on `save_status` is talking about internal implementation and not necessary to understand how `after_completed` works.

```markdown
To help with this, we've added an additional callback called `after_completed`.
It will always run after valid operations have completed, even if there is nothing to save to the database.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    ### Changes to the Query objects
+
+    The `Query` objects also got a few new changes. The main one being that queries no longer mutate the object. Prior to this release, appending any
+    query method would mutate the original object. This made building queries really nice and easy, but presented a problem that many faced. For example:
+
+    ```crystal
+    q = UserQuery.new
+    if current_user.admin?
+      q.admin(true)
+    end
```
matthewmcgarvey:
```suggestion
```

These lines aren't necessary. It took me a few seconds to realize it

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    ```
+
+    In that case, you had to make sure you cloned the query, and they were in the correct order. Now we no longer mutate the query, but this requires changing your
+    code just slightly.
+
+    ```crystal
+    q = UserQuery.new
+    if current_user.admin?
+      q = q.admin(true)
+    end
```
matthewmcgarvey:
```suggestion
```
Same as above

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    user_total_count = q.select_count
+    users = q.username.asc_order
+    ```
+
+    With this change, this also means that defining default queries in your query class `initialize` need to change. To set a default, you'll now use the `defaults` method.
+
+    ```crystal
+    class UserQuery < User::BaseQuery
```
matthewmcgarvey:
```suggestion
    class AdminQuery < User::BaseQuery
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+      def initialize
+        defaults &.admin(true)
+      end
+    end
+    ```
+
+    ### Type-safe WHERE "OR"
+
+    This was one of the most requested methods we got! We held off since adding this in makes SQL queries quite a bit more complex, but we finally got it! (**sort of... see note below)
```
matthewmcgarvey:
```suggestion
    This was one of the most requested additions! We held off since adding this in makes SQL queries quite a bit more complex, but we finally got it! (**sort of... see note below)
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    ```crystal
+    UserQuery.new.name("Billy").or(&.name("Kelly").age.gte(51)).or(&.admin(true))
+    ```
+
+    This query would generate `WHERE name = 'Billy' OR name = 'Kelly' AND age >= 51 OR admin = true`, and that may not give you the result you expect. We are working on a solution,
+    but for now if you need more complex control, you can pass raw SQL to `UserQuery.new.where("name = ? OR (name = ? && age >= ?)", "Billy", "Kelly", 51)`.
+
+    ### Model Association updates
+
+    When you're using a RDBMS like postgres, table associations become very important for properly structuring your data. There were several bugs fixed when it comes to
```
matthewmcgarvey:
```suggestion
    When you're using an RDBMS like PostgreSQL, table associations become very important for properly structuring your data. There were several bugs fixed when it comes to
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    associations giving us a lot more power and control over our data.
```
matthewmcgarvey:
```suggestion
    associations giving you a lot more power and control over your data.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ```crystal
+    class Employee < BaseModel
+      table do
+        belongs_to boss : Manager
+      end
+    end
+    ```
+
+    But this would throw an error due to how the query methods were generated. We would be expecting to generate a method named `where_managers`, but that doesn't exist here.
```
matthewmcgarvey:
```suggestion
    But this would throw an error due to how the query methods were generated.
    In some places we were using the table name to generate code, and in others we were using the association name.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    With this release, we no longer make the assumption on what you're naming your associations. This also means that a small convention must be followed where your `where_` query
```
matthewmcgarvey:
```suggestion
    With this release, we no longer make an assumption about what you're naming your associations. This also means that a small convention must be followed where your `where_` query
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+        has_many posts : Post
+
+        # you specify an Array(Symbol) where the first item is the method to query through
+        # and the second item is that association's method.
+        has_many comments : Comment, through: [:posts, :comments]
+      end
+    end
+    ```
+
+    This means you can also now do a has_many though a has_many through 😬
```
matthewmcgarvey:
```suggestion
    Related to this change, we've also fixed bugs that limited the types of associations that could be used for the "through" association 😬
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ### Models now support SQL VIEW
+
+    Generally when we think of models, we think of database tables. The models even have a `table()` method to denote this!
+    Well now there's a `view()` method as well!
+
+    > SQL VIEWs are like tables, but generally READ-ONLY, and have no primary key. Their data is usually aggregated from other table sources.
```
matthewmcgarvey:
```suggestion
    > SQL VIEWs are like tables, but generally READ-ONLY, and might not have a primary key. Their data is usually aggregated from other table sources.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+      view do
+        column username : String
+        column promoted_to_admin_on : Time
+      end
+    end
+    ```
+
+    All of the columns for your `view` must be manually defined. These models do not come with any sort of `primary_key`, or `id` methods. This also means that methods like
+    `AdminUserQuery.find()`, and `admin_user.delete` won't work since these all rely on there being an `id` method. You must implement any of these methods yourself that you may
+    need.
```
matthewmcgarvey:
```suggestion
    All of the columns for your `view` must be manually defined. These models do not come with any sort of `primary_key` or `timestamps` by default. If your view has these columns, you will need to explicitly add them.
    View models will not have a `SaveOperation` defined as they are meant to be read-only, and if no primary key is added, some features will be missing from the model and `BaseQuery`.
    For example, `AdminUserQuery.find()` and `admin_user.reload` won't work since these rely on an `id` method.
    You must implement any of these methods yourself if you need them.
```

Trying to make it clear that some features could be missing because they don't have a primary_key but others just aren't available (SaveOperation)

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ### Routing Changes
+
+    The LuckyRouter got a few new fancy that we're super stoked about! The first one is "glob" routing.
```
matthewmcgarvey:
```suggestion
    The LuckyRouter got some fancy upgrades that we're super stoked about! The first one is "glob" routing.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    Glob routes are basically a route where you know the first part in the path, but the end of the route path is a variable length. For example:
```
matthewmcgarvey:
```suggestion
    Glob routes are basically a route where you might know some parts of the path, but the end of the route path is a variable length. For example:
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ```
+    myblog.com/posts
+    myblog.com/posts/2020
+    myblog.com/posts/2020/12
+    myblog.com/posts/2020/12/25
+    ```
+
+    These routes may all just display posts, and you don't want to duplicate routes or pages for the same actions being taken. In this case, we can define a glob route
+    to catch all of these in to the same action.
```
matthewmcgarvey, line 300:
```suggestion
    to route all of these in to the same action.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ```crystal
+    class Posts::Index < BrowserAction
+      get "/posts/*:date" do
+        year, month, day = date.split('/')
```
matthewmcgarvey: I don't think this actually works with the urls above... but I don't have a good idea on how to fix it haha

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    They have been huge contributors in the community, and since bringing them on, the movement towards 1.0 has more than doubled in speed!
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    Or you can view the changes using [LuckyDiff](https://luckydiff.com/?from=0.24.0&to=0.25.0).
+    This tool helps see what has changed between versions and will be included in upgrade notes from now on.
+
+    ## Here's what's new
+
+    Since the last release we've welcomed two more developer to the core team, [Matthew](https://github.com/matthewmcgarvey) and [Stephen](https://github.com/stephendolan).
+    These two have both been huge contributors in the commuinity, and since bringing them on, the movement towards 1.0 has more than doubled in speed!
+
+    Over the course of the last few months, we've made a ton of changes for this release, so let's break down a couple of the more notible changes:
```
edwardloveall:
```suggestion
    Over the last few months, we've made a ton of changes, so let's break down a couple of the more notable ones:
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+        html IndexPage, posts: PostQuery.new.by_date(year, month, day)
+      end
+    end
+    ```
+
+    Another great update to the router is that we can now catch duplicate routes defined. This helps you to catch mistakes in development when you're working on a larger
+    code base, and accidentally defined two conflicting routes.
+
+    ```crystal
+    # These will now raise an error
```
matthewmcgarvey:
```suggestion
    # These will now raise an error when starting the app
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    ## Here's what's new
+
+    Since the last release we've welcomed two more developer to the core team, [Matthew](https://github.com/matthewmcgarvey) and [Stephen](https://github.com/stephendolan).
+    These two have both been huge contributors in the commuinity, and since bringing them on, the movement towards 1.0 has more than doubled in speed!
+
+    Over the course of the last few months, we've made a ton of changes for this release, so let's break down a couple of the more notible changes:
+
+    ### `Avram::Operation` got a facelift
+
+    When you need to handle some logic that may not be tied directly to a single model, you can create an `Avram::Operation` to handle this. You've probably seen
```
edwardloveall:
```suggestion
    When you need to handle some logic that isn't tied directly to a single model, you can create an `Avram::Operation` to handle this. You've probably seen
```

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    examples of this in the default generated app like `RequestPasswordReset` and `SignInUser`.
```

matthewmcgarvey:
Feel free to ignore any suggested changes you don't like. I'm no English major so I most just go by how it sounds to me

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    examples in your app like the `RequestPasswordReset` and `SignInUser` that come with an Auth app.
+
+    Prior to this update, it was sort of "wild west", and completely up to you on how to implement these. For consistency sake, we recommended defining a `submit`
+    method, and then returning `yield self, value`. These operations were also limited with no way to use callbacks, file attributes, or define errors not tied to
```
edwardloveall: Just a point of clarity, it is literally: is `yield self, value` the literal syntax? Is `yield` method that expects multiple arguments? That's cool if so, I just wasn't aware of it 😄

edwardloveall:
```suggestion
    method, and then returning `yield self, value`. These operations were also limited as they couldn't use callbacks, file attributes, or define errors not tied to
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    an attribute.
+
+    In this release, we've created a whole new interface!
```
edwardloveall, line 44: After reading the code block, I'm not sure what's new. Is it the `def run`? It might be good to highlight specifically what's new and worth getting excited about.

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+        # after save
+      end
+
+      after_commit do |saved_user|
+        # after DB commit
+      end
+    end
+    ```
+
+    We have also added the ability to do these callbacks conditionally. This allows you to specify if the callbacks should actually run or not.
```
edwardloveall:
```suggestion
    We've also added the ability to conditionally trigger these callbacks based on a method.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ```crystal
+    class SaveUser < User::SaveOperation
```
edwardloveall, line 113: Great example!

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+      private def update_billing
+      end
+
+      private def card_still_valid?
+        false
+      end
+    end
+    ```
+
+    Lastly, there were some instances where you may be updating a record, and expecting a callback when everything was done. However, in the case of
```
edwardloveall:
```suggestion
    And finally for callbacks, when an operation doesn't update a record, the `after_save` and `after_commit` callbacks aren't triggered.
```

I'd just delete the next sentence and combine it with the next paragraph.

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    We've added one additional callback called `after_completed`. This callback is always called when the operation is successful (the internal `save_status` attribute is set to `:saved`).
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ### `with_defaults` is now `tag_defaults`
+
+    When you create components like `Shared::Field`, you can use the `tag_defaults` method to apply default attributes to a set of HTML elements.
+    This method was called `with_defaults` before, but has been renamed for some clarity on what it does.
+
+    ### Changes to the Query objects
+
+    The `Query` objects also got a few new changes. The main one being that queries no longer mutate the object. Prior to this release, appending any
+    query method would mutate the original object. This made building queries really nice and easy, but presented a problem that many faced. For example:
```
edwardloveall:
```suggestion
    query method would mutate the original object. This made building queries nice and easy, but presented a problem that many faced. For example:
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+      q.admin(true)
+    end
+
+    # this would fail because we've already mutated the query
+    # on the count, and can no longer query with the username ordering
+    user_total_count = q.select_count
+    users = q.username.asc_order
+    ```
+
+    In that case, you had to make sure you cloned the query, and they were in the correct order. Now we no longer mutate the query, but this requires changing your
```
edwardloveall:
```suggestion
    In that case, you had to make sure you cloned the query, and the chained methods were called in the correct order. We no longer mutate the query, but this requires a code change.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    code just slightly.
```
edwardloveall:
```suggestion
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ```crystal
+    q = UserQuery.new
+    if current_user.admin?
+      q = q.admin(true)
```
edwardloveall:
```suggestion
      q = q.admin(true) # mutate the query explicitly
```

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    This was one of our most common requests! We held off since adding this in makes SQL queries quite a bit more complex, but we finally got it! (**sort of... see note below)
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    ### Type-safe WHERE "OR"
+
+    This was one of the most requested methods we got! We held off since adding this in makes SQL queries quite a bit more complex, but we finally got it! (**sort of... see note below)
+
+    ```crystal
+    # WHERE users.name = 'Billy' OR users.name = 'Kelly'
+    UserQuery.new.name("Billy").or(&.name("Kelly"))
+    ```
+
+    As you can see, you call the `or` method which passes in the instance of the query object allowing you to chain additional WHERE clauses.
```
edwardloveall:
```suggestion
    Calling the `or` method which passes in the instance of the query object allowing you to chain additional WHERE clauses.
```

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    When using Postgres (Avram's RDBMS), table associations become a critical part of structuring your data. There were several bugs fixed when it comes to
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    This query would generate `WHERE name = 'Billy' OR name = 'Kelly' AND age >= 51 OR admin = true`, and that may not give you the result you expect. We are working on a solution,
+    but for now if you need more complex control, you can pass raw SQL to `UserQuery.new.where("name = ? OR (name = ? && age >= ?)", "Billy", "Kelly", 51)`.
+
+    ### Model Association updates
+
+    When you're using a RDBMS like postgres, table associations become very important for properly structuring your data. There were several bugs fixed when it comes to
+    associations giving us a lot more power and control over our data.
+
+    To start, we will look at the `belongs_to` association method. It's pretty common to name your method the same as the model it's referencing, but in some cases,
```
edwardloveall:
```suggestion
    To start, let's look at the `belongs_to` association method. It's pretty common for the method name to be the same as the model it's referencing, but in some cases,
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    you may want your association method to be named slightly different like this:
```
edwardloveall:
```suggestion
    you may want your association method to be named something different:
```

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    With this release, we no longer make the assumption on what you're naming your associations. But it's also worth pointing out that `where_` query
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    But this would throw an error due to how the query methods were generated. We would be expecting to generate a method named `where_managers`, but that doesn't exist here.
+
+    With this release, we no longer make the assumption on what you're naming your associations. This also means that a small convention must be followed where your `where_` query
+    methods will append the name of the association.
+
+    ```crystal
+    EmployeeQuery.new.where_boss(ManagerQuery.new)
+    ```
+
+    To go along with the `belongs_to`, we've also fixed some bugs related to using `has_many through`. This update will require a slightly different syntax.
```
edwardloveall:
```suggestion
    We've also fixed some bugs related to using `has_many through`. This update will require a slightly different syntax.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    end
+
+    # After update
+    class User < BaseModel
+      table do
+        has_many posts : Post
+
+        # you specify an Array(Symbol) where the first item is the method to query through
+        # and the second item is that association's method.
+        has_many comments : Comment, through: [:posts, :comments]
```
edwardloveall, line 258:
```suggestion
        has_many comments : Response, through: [:posts, :responses]
```

The thought here is that you'd need to name this one because it's different than the association name. Is this example good enough?

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+      end
+    end
+    ```
+
+    This means you can also now do a has_many though a has_many through 😬
+
+    ### Models now support SQL VIEW
+
+    Generally when we think of models, we think of database tables. The models even have a `table()` method to denote this!
+    Well now there's a `view()` method as well!
```
edwardloveall, line 268: Ok but this is RAD!

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+        has_many comments : Comment, through: [:posts, :comments]
+      end
+    end
+    ```
+
+    This means you can also now do a has_many though a has_many through 😬
+
+    ### Models now support SQL VIEW
+
+    Generally when we think of models, we think of database tables. The models even have a `table()` method to denote this!
```
edwardloveall:
```suggestion
    Generally when we think of models, we think of database tables. The models even have a `table()` method to denote this.
```

Slow it down, partner 😂

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    All of the columns for your `view` must be manually defined. These models do not come with any sort of `primary_key`, or `id` methods unless you add them. This also means that methods like
    `AdminUserQuery.find()`, and `admin_user.delete` won't work by default since they rely on there being an `id` method. You must implement any of these methods yourself that you may
    need.
```

src/posts/lucky_0.25.cr
edwardloveall:
```suggestion
    Glob routes are a route where the first part in the path is known, but the end of the route path is a variable length. For example:
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    Glob routes are basically a route where you know the first part in the path, but the end of the route path is a variable length. For example:
+
+    ```
+    myblog.com/posts
+    myblog.com/posts/2020
+    myblog.com/posts/2020/12
+    myblog.com/posts/2020/12/25
+    ```
+
+    These routes may all just display posts, and you don't want to duplicate routes or pages for the same actions being taken. In this case, we can define a glob route
```
edwardloveall:
```suggestion
    If these routes all display posts, it's not useful to duplicate routes or pages over multiple actions. In this case, we can define a glob route
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+      get "/posts/*:date" do
+        year, month, day = date.split('/')
+
+        html IndexPage, posts: PostQuery.new.by_date(year, month, day)
+      end
+    end
+    ```
+
+    Another great update to the router is that we can now catch duplicate routes defined. This helps you to catch mistakes in development when you're working on a larger
+    code base, and accidentally defined two conflicting routes.
```
edwardloveall:
```suggestion
    Another great update to the router is that Lucky can now catch overridden (duplicate) routes. This helps catch mistakes in development, especially in large project where hundreds or more routes are defined.
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    end
+
+    class Api::Posts::SearchByYear < ApiAction
+      get "/api/posts/:year" do
+        #...
+      end
+    end
+    ```
+
+    ### So many more!
```
edwardloveall:
```suggestion
    ### And much more!
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    This blog post could go on for days to include all the awesome stuff we've added. Here's a super quick run down of a few more:
```
edwardloveall:
```suggestion
    This blog post could go on for days to include all the awesome stuff we've added. Here's a quick run down of a few more:
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    * Updates and new features to LuckyFlow like hover over element
+    * More bug fixes in associations and file handling
+    * Global use of `memoize` in any class
+    * Testing CLI Tasks is easier
+    * Lots of cleanup, and refactors
+
+    Read through the [CHANGELOG](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md) to see what else!
```
edwardloveall:
```suggestion
    Read through the [CHANGELOG](https://github.com/luckyframework/lucky/blob/master/CHANGELOG.md) to see it all!
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    ## Parting words
+
+    This is another step towards 1.0, and we're super stoked. We can't do this without continued support from our community.
+    The more hands and eyes on the project, the more fine tuned we can make this project!
```
edwardloveall:
```suggestion
    The more hands and eyes on the project, the more fine tuned Lucky can be!
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+
+    Please give it a spin and help us find bugs so our next release is even more solid.
+    If you find any issues, don't hesitate to [report the issue](https://github.com/luckyframework/lucky/issues).
```
edwardloveall:
```suggestion
    If you find any issues, don't hesitate to [report them](https://github.com/luckyframework/lucky/issues).
```

src/posts/lucky_0.25.cr
```diff
@@ -0,0 +1,370 @@
...
+    If you're unsure, just hop on [Discord chat](https://discord.gg/HeqJUcb) so we can help you out.
+
+    Thanks so much for the support!
+
+    ### Follow and spread the word
+
+    If you haven't already, give us a [star on github](https://github.com/luckyframework/lucky),
```
edwardloveall:
```suggestion
    If you haven't already, give us a [star on GitHub](https://github.com/luckyframework/lucky),
```

src/posts/lucky_0.25.cr
edwardloveall: I was trying (slightly) to avoid `us` or `you` just making it person-less. But I do like `you` over `us`.

src/posts/lucky_0.25.cr
edwardloveall: I'm a little hesitant to use `route` as a verb and a noun, even though I see where you're coming from here.

src/posts/lucky_0.25.cr
edwardloveall: Yeah this is tricky. I was thinking about if we could type these routes somehow? But it might just be an unsafe thing we do unless we have users build parser combinators which is a bit more haskell than I like in my web frameworks.

src/posts/lucky_0.25.cr
jwoertink: apparently I can't spell that word 😂

jwoertink: oh, c'mon github, really? Because I commented on this, I can't add that change to the batch. Lame

jwoertink: I was wrong... It's because that suggestion has more than 1 change in it. 🤨

src/posts/lucky_0.25.cr
jwoertink: yet this one has multiple changes 🤔

src/posts/lucky_0.25.cr
jwoertink: These aren't compile-time, right? They're runtime errors? When I was fixing the Lucky repo, specs were running when I'd get an error.

src/posts/lucky_0.25.cr
jwoertink: Yeah, you have to define `run` now. Before you could have called the method whatever you wanted because it was up to you to use whatever you defined.

src/posts/lucky_0.25.cr
jwoertink: Whether you name it different or not, you have to use this new syntax. I guess we could mention that it allows you to name things differently if you need 🤔

src/posts/lucky_0.25.cr
jwoertink: > the `after_save` and `after_commit` callbacks aren't triggered.

This has always been the case, so that's not really different. Maybe I just get rid of that whole line?

src/posts/lucky_0.25.cr
jwoertink: Ok, so it does work, but needs a slight adjustment..

```crystal
    if date
      year, month, day = date.not_nil!.split('/') + [nil, nil]
    end
```

Would this be too confusing?

matthewmcgarvey (approved):
❤️

PR #520, opened 2020-12-19
Fixing an inccorect setup for alternate ORMs. 

### LINKED ISSUE #496: Incorrect documentation in alternate ORM guide
https://luckyframework.org/guides/database/intro-to-avram-and-orms#alternate-orms

The configuration no longer takes `"unused"`, instead we would set `Avram::Credentials.void`

jwoertink:
Fixes #496

### DIFF
--- a/src/actions/guides/database/intro_to_avram_and_orms.cr
+++ b/src/actions/guides/database/intro_to_avram_and_orms.cr
@@ -107,7 +107,7 @@ class Guides::Database::IntroToAvramAndORMs < GuideAction
     * Be sure to include the adapter in your `shard.yml` file
     * Still place your models in the `src/models` directory
     * `src/app_database.cr` can be deleted to avoid confusion
-    * Avram must still be configured, but can be set as "unused"
+    * Avram must still be configured, but can be set as "unused" (see below)
     * Update your `config/database.cr` file with the new adapter's configuration settings
 
     ```crystal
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/intro_to_avram_and_orms.cr
```diff
@@ -125,15 +125,16 @@ class Guides::Database::IntroToAvramAndORMs < GuideAction
...
     end
     ```
 
     * Place migrations (if necessary) in `db/migrations/`.
     * Boxes, and Queries are specific to Avram Models, but you can still use [Basic Operations](#{Guides::Database::ValidatingSaving.path}).
 
-    > If your app doesn't need a database, you should still set the `AppDatabase` configure setting to
-    > some non-empty string. Avram Operations can still be quite useful for things like contact forms,
+    > If your app doesn't need a database, you should still set the `AppDatabase` credentials setting to
+    > `Avram::Credentials.void`. Avram Operations can still be quite useful for things like contact forms,
```
matthewmcgarvey, line 137: Are these lines needed since we have explicit docs above?

matthewmcgarvey (approved):

src/actions/guides/database/intro_to_avram_and_orms.cr
jwoertink: just a little extra read for people skimming the docs and not actually reading, I guess.

PR #522, opened 2020-12-19
Add some docs on settings queries with defaults. 

### LINKED ISSUE #512: Document new way to set default queries
https://github.com/luckyframework/avram/pull/564

jwoertink:
Fixes #512


matthewmcgarvey (approved):
You might also mention or point to how to "undo" the queries added in the defaults if necessary. (Like when the default is ignoring soft-deleted records).  Is there just a way to "clear" the queries list?

jwoertink:
Yeah, I'm pretty sure there's a way to clear the default and reset it back... I don't think it's straight forward, but I'll take a look at that.

PR #527, opened 2020-12-21
Updated guides on associations

### LINKED ISSUE #483: Document association options
The model association methods like `belongs_to` have options like `foreign_key`, or `table` that should be documented. 

https://github.com/luckyframework/avram/blob/master/src/avram/associations/belongs_to.cr#L2
https://luckyframework.org/guides/database/models#belongs-to

jwoertink:
Fixes #483

### DIFF
--- a/src/actions/guides/database/models.cr
+++ b/src/actions/guides/database/models.cr
@@ -424,7 +424,7 @@ class Guides::Database::Models < GuideAction
       table do
         column name : String
         has_many taggings : Tagging
-        has_many posts : Post, through: :taggings
+        has_many posts : Post, through: [:taggings, :post]
       end
     end
 
--- a/src/actions/guides/database/models.cr
+++ b/src/actions/guides/database/models.cr
@@ -424,6 +424,12 @@ class Guides::Database::Models < GuideAction
       table do
         column name : String
         has_many taggings : Tagging
+
+        # In the has_many :through example below, the `:taggings`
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/models.cr
```diff
@@ -407,19 +424,22 @@ class Guides::Database::Models < GuideAction
...
       table do
         column title : String
         has_many taggings : Tagging
-        has_many tags : Tag, through: :taggings
+        has_many tags : Tag, through: [:taggings, :tag]
       end
     end
     ```
 
+    The `through` is always set to an array like in the example `[:taggings, :tag]`. The first item `:taggings` is a Symbol name
```
jwoertink: This was actually kind of hard to explain. I'm hoping it just "makes sense" to people, but if anyone has a better way to word this without getting "too wordy", I'm open.

matthewmcgarvey: You might read through Ecto's docs on this and see if anything they have is more clear https://hexdocs.pm/ecto/Ecto.Schema.html#has_many/3-has_many-has_one-through

matthewmcgarvey (approved):

src/actions/guides/database/models.cr
jwoertink: Ok, I updated that. Just literally copied what they had because that was pretty clear. It's a bit more wordy than I wanted, but as long as people get it, that's all that really matters.

PR #532, opened 2020-12-21
Adding docs on the cycle page helper method.

### LINKED ISSUE #487: Document cycle text helper
https://github.com/luckyframework/lucky/blob/master/src/lucky/page_helpers/text_helpers.cr#L160

Should go here https://luckyframework.org/guides/frontend/rendering-html#formatting-helpers

jwoertink:
Fixes #487

### DIFF
--- a/src/actions/guides/frontend/rendering_html.cr
+++ b/src/actions/guides/frontend/rendering_html.cr
@@ -586,17 +586,14 @@ class Guides::Frontend::RenderingHtml < GuideAction
 
     ```crystal
     posts.each do |post|
-      tr class: cycle(["bg-gray-600", ""], "") do
+      tr class: cycle(["bg-gray-600", ""]) do
         td post.title
       end
     end
     ```
 
-    In this example, the first row, and all even rows, will be `<tr class="bg-gray-600">`, but the next row, and all odd rows, will be
+    In this example, the first row, and all odd rows, will be `<tr class="bg-gray-600">`, but the next row, and all even rows, will be
     `<tr class="">`. You can pass as many values as you'd like to cycle through on each iteration.
-
-    The second argument to the `cycle` method is the default value you would like to start with. If no value is given, the string `"default"`
-    is assumed.
 
     ## Layouts
 
--- a/src/actions/guides/frontend/rendering_html.cr
+++ b/src/actions/guides/frontend/rendering_html.cr
@@ -448,7 +448,7 @@ class Guides::Frontend::RenderingHtml < GuideAction
     # => true
     ```
 
-    ## Formatting helpers
+    ## Formatting and Page helpers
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/frontend/rendering_html.cr
```diff
@@ -579,6 +579,25 @@ class Guides::Frontend::RenderingHtml < GuideAction
...
+
+    ```crystal
+    posts.each do |post|
+      tr class: cycle(["bg-gray-600", ""], "") do
+        td post.title
+      end
+    end
+    ```
+
+    In this example, the first row, and all even rows, will be `<tr class="bg-gray-600">`, but the next row, and all odd rows, will be
```
matthewmcgarvey:
```suggestion
    In this example, the first row, and all odd rows, will be `<tr class="bg-gray-600">`, but the second row, and all even rows, will be
```
Not sure how it works, but I think this is correct?
1 - "bg-gray-600"
2 - ""
3 - "bg-gray-600"
4 - ""

src/actions/guides/frontend/rendering_html.cr
```diff
@@ -579,6 +579,25 @@ class Guides::Frontend::RenderingHtml < GuideAction
...
+    `<tr class="">`. You can pass as many values as you'd like to cycle through on each iteration.
+
+    The second argument to the `cycle` method is the default value you would like to start with. If no value is given, the string `"default"`
```
matthewmcgarvey: I don't understand this at all. Maybe you could combine all the explanations and then update the code example to also show what it translates into?

src/actions/guides/frontend/rendering_html.cr
jwoertink: Yeah, that's correct. I guess I was basing it off the index (starting with 0).... 
0 - "bg-gray-600"
1 - ""
2 - "bg-gray-600"
3 - ""

😅  very programmer of me. But maybe your suggestion makes more sense

src/actions/guides/frontend/rendering_html.cr
jwoertink: I actually don't understand how that works either. The whole cycle system is a bit complex. My guess is it's something like this (untested)

```crystal
cycle([1, :cheese, "something", 42])

1 - "default"
3 - "1"
4 - "cheese"
5 - "something"
6 - "42"
```

because the "default" value for that second argument is `"default"`. It's kind of confusing though, and in the example of alternating classes, I don't think I've ever used a classname of "default" before 😕  Maybe I just leave that off, and add "default" in to the example?

jwoertink: ok, actually.. I just threw it in to a fresh new app to test, and I have no clue what that second argument does. It didn't seem to matter what I put in it... So I'm just going to leave it off and pretend it doesn't exist for now 😂

matthewmcgarvey (approved):

PR #533, opened 2020-12-21
Add docs to mention use of UPPER/LOWER in queries.

### LINKED ISSUE #484: Document String LOWER() and UPPER()
https://github.com/luckyframework/avram/blob/master/src/avram/charms/string_extensions.cr#L43

If you need something like `SELECT * FROM users WHERE LOWER(username) = ?` you would do

```crystal
UserQuery.new.username.lower.eq("whatever")
```

Same for upper.

This would go in the query guide https://luckyframework.org/guides/database/querying-records#where-queries

jwoertink:
Fixes #484

### DIFF
--- a/src/actions/guides/database/querying.cr
+++ b/src/actions/guides/database/querying.cr
@@ -236,6 +236,22 @@ class Guides::Database::Querying < GuideAction
     UserQuery.new.name.is_not_nil
     ```
 
+    ### LOWER/UPPER A = B
+
+    Find rows (casting `A` to LOWER/UPPER) is equal to `B`
+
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/querying.cr
```diff
@@ -236,6 +236,22 @@ class Guides::Database::Querying < GuideAction
     UserQuery.new.name.is_not_nil
     ```
 
+    ### LOWER/UPPER A = B
+
+    Find rows (casting `A` to LOWER/UPPER) is equal to `B`
```
matthewmcgarvey:
```suggestion
    Find rows where (casting `A` to LOWER/UPPER) is equal to `B`
```

matthewmcgarvey (approved):

PR #540, opened 2020-12-22
Added new guide section for handling files. This starts with file uploads

### LINKED ISSUE #84: mention file/media uploads
The docs could provide some info about file uploads.

(Ideally providing a versioned download filename, and a default path that always redirects to latest replacement version.)

### LINKED ISSUE #330: Document file uploads
As described in https://github.com/luckyframework/lucky/issues/1194.

jwoertink:
Fixes #330
Fixes #84 

This is just a quick guide so we have something. My direction for this new guide section will be that we can have a guide on file downloads (sending files to client), as well as more stuff on how to test files, and more on the actual Lucky file object. For now, this is more than what we have and should get people going.

Thanks @matthewmcgarvey for writing up the example.

### DIFF
--- a/src/actions/guides/handling_files/file_uploads.cr
+++ b/src/actions/guides/handling_files/file_uploads.cr
@@ -1,0 +1,104 @@
+class Guides::HandlingFiles::FileUploads < GuideAction
+  guide_route "/handling-files/file-uploads"
+
+  def self.title
+    "Uploading Files"
+  end
+
+  def markdown : String
+    <<-MD
+    ## Setup
+
+    For handling file uploads, we will use [Shrine.cr](https://github.com/jetrockets/shrine.cr).
+
+    Start by adding the `shrine` shard to your `shard.yml` and run `shards install`.
+
+    ```yaml
+    dependencies:
+      shrine:
+        github: jetrockets/shrine.cr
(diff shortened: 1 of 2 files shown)


src/actions/guides/handling_files/file_uploads.cr
```diff
@@ -0,0 +1,104 @@
...
+      permit_columns name
+      file_attribute :profile_picture
+
+      before_save do
+        profile_picture.value.try { |pic| upload_pic(pic) }
+      end
+
+      private def upload_pic(pic)
+        result = Shrine.upload(File.read(pic.tempfile.path), "store", metadata: { "filename" => pic.filename })
+        profile_picture_path.value = result.url
```
matthewmcgarvey:
```suggestion
        profile_picture_path.value = result.id
```

Storing the `id` instead of the `url` allows you to use `Shrine` later (like if you want to delete the uploaded file).  Shrine's docs actually have you turn the whole `result` into json and store that but 🤷

matthewmcgarvey (approved):
Minor tweak. I wrote the docs after the first time I added Shrine to a project. I have changed some stuff the second time around for the better but I don't want to complicate this guide with the other changes.

PR #560, opened 2021-02-01
Update for buildpack target change

robacarp:
Updates lucky website for https://github.com/luckyframework/heroku-buildpack-crystal/pull/11

I also removed `pbcopy` which I think is a result of someone typing `cat .github/actions/something > pbcopy` instead of `| pbcopy`... though I suppose it's possible [that it's there on purpose](https://github.com/robacarp/mosquito/blob/master/:w)


jwoertink (approved):
LOL `pbcopy` 😂 

This is fine, but let's wait on the other two PRs to be merged in first before doing this.

PR #579, opened 2021-02-07
Import only the highlight.js core.

usern3:
The entire highlight.js was being imported, so the lines below importing the individual langs were unnecessary. 

With this change it only imports the core and the languages you have imported below it rather than all of the langs supported by the library, which seems like was the intention.

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -9,7 +9,7 @@ import Turbolinks from "turbolinks";
 RailsUjs.start();
 Turbolinks.start();
 
-import hljs from 'highlight.js';
+import hljs from 'highlight.js/lib/core';
 
 import javascript from 'highlight.js/lib/languages/javascript';
 hljs.registerLanguage('javascript', javascript);


jwoertink (approved):
WHOA!!!.... hold up.....  I just tested this locally...

```
BEFORE
/js/app.js │ 1.45 MiB

AFTER
/js/app.js │ 194 KiB
```

🤯  That's a bonkers amount of performance gain 🤣

usern3:
> WHOA!!!.... hold up.....  I just tested this locally...
> 
> ```
> BEFORE
> /js/app.js │ 1.45 MiB
> 
> AFTER
> /js/app.js │ 194 KiB
> ```
> 
> 🤯  That's a bonkers amount of performance gain 🤣 

Oh wow! More than I though for sure. That's nearly a factor of 10 haha. It might've been doubling up on some of the code because of the following imports. JS isn't my strong point at all so I'm not sure if it works like that, but maybe. 😝

PR #587, opened 2021-02-08
Added HTML to fix broken highlighting on the HTML section

usern3:
Currently highlighting is broken in the HTML sections of the website. This _should_ fix it.

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -32,9 +32,6 @@ hljs.registerLanguage('sql', sql);
 import erb from 'highlight.js/lib/languages/erb';
 hljs.registerLanguage('erb', erb);
 
-import xml from 'highlight.js/lib/languages/xml';
-hljs.registerLanguage('xml', xml);
-
 import html from 'highlight.js/lib/languages/xml';
 hljs.registerLanguage('html', html);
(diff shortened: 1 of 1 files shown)


src/js/app.js
```diff
@@ -32,6 +32,12 @@ hljs.registerLanguage('sql', sql);
 import erb from 'highlight.js/lib/languages/erb';
 hljs.registerLanguage('erb', erb);
 
+import xml from 'highlight.js/lib/languages/xml';
+hljs.registerLanguage('xml', xml);
```
jwoertink: I don't see any xml code blocks in the guides. We can probably skip this, right?

jwoertink (commented):
Thanks for catching this. Just one question about the xml. I think the html import is fine... doesn't look like highlight distinguishes between `html` and `xml` highlighting.

src/js/app.js
usern3: I included it because I wasn't sure how the plugin handled it because it seems to include HTML and XML together in the same file. I think we should be safe removing it.

jwoertink (approved):
Sweet! Thanks

PR #618, opened 2021-04-10
Announcing Lucky 0.27

jwoertink:
Adding a new blog post for Lucky 0.27


src/posts/lucky_0.27.cr
```diff
@@ -0,0 +1,104 @@
...
+
+    Or you can view the changes using [LuckyDiff](https://luckydiff.com/?from=0.26.0&to=0.27.0).
+    This tool helps see what has changed between versions and will be included in upgrade notes from now on.
+
+    ## Here's what's new
+
+    ### Crystal 1.0 Support
+
+    For the most part, Crystal 1.0 worked with Lucky 0.26 with the exception of cookies.
+    For the Crystal 1.0 release, there was a few changes around cookies:
```
matthewmcgarvey:
```suggestion
    For the Crystal 1.0 release, there were a few changes around cookies:
```

matthewmcgarvey (approved):

src/posts/lucky_0.27.cr
```diff
@@ -0,0 +1,104 @@
...
+    <<-TEXT
+    Crystal 1.0.0 is out! 🥳 And Lucky
+    0.27.0 is fully compatible with it.
+    TEXT
+  end
+
+  def content : String
+    <<-MD
+    We want to first congratulate the Crystal team for their long awaited release of
+    1.0! This is a huge milestone for not only the language, but the whole community.
```
edwardloveall: Thoughts on linking to their blog post?

src/posts/lucky_0.27.cr
```diff
@@ -0,0 +1,104 @@
...
+
+    We're sure the next question on your mind is "When will Lucky hit 1.0?". Don't fret!
+    We're working towards that, and with each release we get closer and closer.
```
edwardloveall, line 23: May be worth coming out and saying here what we said on Discord, that it won't be for the next couple releases, at least. It'll be out when it's ready. etc. It's pinned in Discord if you want wording.

src/posts/lucky_0.27.cr
```diff
@@ -0,0 +1,104 @@
...
+
+    ### LuckyTask
+
+    This was originally called `LuckyCli::Task` which existed in the [lucky_cli](https://github.com/luckyframework/lucky_cli) shard.
+    LuckyCli was required in any shard that included custom tasks. (e.g. Avram, Lucky, Breeze, etc...)
+
+    By abstracting this out to a separate shard, we gain about ~17% (or more) speed in compilation
+    time. This is mainly on cold cache, but it means your apps won't sit in the CI as long as they were!
+
+    Who doesn't love some performance boost?
```
edwardloveall: Technically it'd call it a "speed up" because it doesn't affect (I think) how an app performs, per se. I put this here so that if you are worried about that kind of distinction you can change it, but don't worry about it if you're not 😄

src/posts/lucky_0.27.cr
jwoertink: Good idea!

src/posts/lucky_0.27.cr
jwoertink: Cool. I re-worded it slightly. Let me know if it sounds ok. I don't want it to come off as "Stand off-ish" like "don't bug us about it". But instead more like "in reality, this is the true reason.."

edwardloveall: I like it! I think you give a good reason that doesn't sound like we're annoyed at people for asking.

PR #622, opened 2021-04-11
Updating the Shrine docs

### LINKED ISSUE #607: Updates to Uploading FIles docs
There's some lack of clarity and missing docs on https://luckyframework.org/guides/handling-files/file-uploads that can cause users to become stuck or confused.

Needs:

- `User#profile_picture_path` should be changed to `User#profile_image_id` (referenced in the `SaveUser` operation) as the value is not a file/url path but just an id to be given to Shrine and then get the path from there
- There's no docs on how to render the image in the HTML. Get the image path using `Shrine.find_storage("store").url(user.profile_image_id)` and use it with an `img` tag
- Deleting old files
```crystal
  if old_image = path_attribute.original_value
    storage = Shrine.find_storage("store")
    if storage.exists?(old_image)
      storage.delete(old_image)
    end
  end
```
Not the best code, just an example

Related to https://github.com/luckyframework/website/issues/606

jwoertink:
Fixes #607

### DIFF
--- a/src/actions/guides/handling_files/file_uploads.cr
+++ b/src/actions/guides/handling_files/file_uploads.cr
@@ -77,16 +77,17 @@ class Guides::HandlingFiles::FileUploads < GuideAction
         # If multiple models can share an image, run a query before deleting
         # to ensure you're not breaking any references.
         if old_image = profile_image_id.original_value
-          storage = Shrine.find_storage("store")
-          if storage.exists?(old_image)
-            storage.delete(old_image)
-          end
+          delete_old_profile_image(old_image)
         end
 
         profile_image_id.value = result.id
       end
-      private def delete_old_profile_picture(image_id)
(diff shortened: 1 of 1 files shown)


src/actions/guides/handling_files/file_uploads.cr
```diff
@@ -58,13 +72,18 @@ class Guides::HandlingFiles::FileUploads < GuideAction
...
+          storage = Shrine.find_storage("store")
+          if storage.exists?(old_image)
+            storage.delete(old_image)
+          end
         end
-        profile_picture_path.value = result.id
+
+        profile_image_id.value = result.id
       end
       private def delete_old_profile_picture(image_id)
```
matthewmcgarvey: This can be removed

jwoertink: oops 😅  Actually, I'm gonna use that.

matthewmcgarvey (approved):

PR #638, opened 2021-04-20
Breaking callbacks and validations out in to a separate guide.

### LINKED ISSUE #623: Document custom messages on validations
The `validates_uniqueness_of` validation has a default error messages of "is already taken" https://github.com/luckyframework/avram/blob/master/src/avram/database_validations.cr#L25

We need to document that you can pass in a custom error message to this, as well as most of the others here https://github.com/luckyframework/avram/blob/master/src/avram/validations.cr

jwoertink:
Fixes #623

The "Saving and Validating" guide has become pretty huge. This one is a bit tough because there's so many moving parts. I took just the callbacks and validations out and moved them to their own guide. Then renamed the Saving guide to be "Saving and Updating".

Here's what the sidebar will look like now.

Screen Shot 2021-04-20 at 8 15 29 AM

It seems like a large PR, but for the most part, it was copy/paste from what was existing in the current guide. I did update a few outdated things.


src/actions/guides/database/callbacks_and_validations.cr
```diff
@@ -0,0 +1,320 @@
...
+          name.add_error "Mmmyyaahhh!"
+        end
+      end
+    end
+    ```
+
+    ### Custom error messages
+
+    When a validation fails, an error message will be applied to the attribute automatically. For example,
+    the `validate_required` will return `"is required"`, and the `validate_acceptance_of` will return `"must be accepted"`.
```
jwoertink, line 200: This is the section that fixes #623

src/actions/guides/database/callbacks_and_validations.cr
```diff
@@ -0,0 +1,320 @@
...
+    You can pass a `message` option to any of the attributes to define your own custom error message.
+
+    ```crystal
+    before_save do
+      validate_required name, message: "se requiere"
+      validate_confirmation_of email, with: email_confirmation, message: "debe coincidir con"
+    end
+    ```
+
+    ## What are attributes?
```
jwoertink, line 211: This section is technically duplicated between the two guides; however, I think it's fine since you need to know about attributes when doing validations, but you also need to know about them when talking about `attribute`, `permit_columns`, using `changed?`,etc... We could break them out in to their own guide later.

matthewmcgarvey (approved):

PR #649, opened 2021-05-13
Updated return type in documentation for select_count

### LINKED ISSUE #592: Guides say select_count returns Float64... it doesn't
The guides for the `select_count` method [here](https://luckyframework.org/guides/database/querying-records#select-count) say that it returns `Float64`, but it actually returns `Int64` as shown [here](https://github.com/luckyframework/avram/blob/master/src/avram/queryable.cr#L201)

nodeMD:
That PR is addressing Issue 592: https://github.com/luckyframework/website/issues/592

### DIFF
--- a/src/actions/guides/database/querying.cr
+++ b/src/actions/guides/database/querying.cr
@@ -415,7 +415,7 @@ class Guides::Database::Querying < GuideAction
     `SELECT COUNT(*) FROM users`
 
     ```crystal
-    # This will return a Int64.
+    # This will return an Int64.
     # The value will be 0 if there are no records.
     UserQuery.new.select_count
     ```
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
🎉

src/actions/guides/database/querying.cr
```diff
@@ -415,7 +415,7 @@ class Guides::Database::Querying < GuideAction
     `SELECT COUNT(*) FROM users`
 
     ```crystal
-    # This will return a Float64.
+    # This will return a Int64.
```
stephendolan:
```suggestion
    # This will return an Int64.
```

stephendolan (approved):

PR #651, opened 2021-05-20
updated WHERE OR docs with new wrap method.

### LINKED ISSUE #647: Update docs on WHERE OR
https://luckyframework.org/guides/database/querying-records#where-with-or-a-b-or-a-c

We now have a way to place parenthesis with the `where(&)` method. Need to remove the note on this, and add docs on how to use the new method.

jwoertink:
Fixes #647

### DIFF
--- a/src/actions/guides/database/querying.cr
+++ b/src/actions/guides/database/querying.cr
@@ -216,8 +216,19 @@ class Guides::Database::Querying < GuideAction
     UserQuery.new.name("Alfred").or(&.name("Bruce"))
     ```
 
-    > `OR` queries can become quite complex. Avram currently makes no assumption on
-    > where to place parenthesis `()` for conditional grouping. [See #488](https://github.com/luckyframework/avram/issues/488)
+    `OR` queries can become quite complex. If you need to wrap conditions, you can use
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/querying.cr
```diff
@@ -216,8 +216,19 @@ class Guides::Database::Querying < GuideAction
...
+    parenthesis `()`.
+
+    `SELECT COLUMNS FROM users WHERE users.likes_bats = true OR (users.first_name = 'Kate' AND users.last_name = 'Kane')`
+
+    ```crystal
+    UserQuery.new.likes_bats(true).or do |or|
+      or.where do |where|
+        where.first_name("Kate").last_name("Kane"))
+      end
+    end
```
matthewmcgarvey, line 230: Something about the blocks and variable naming bothers me, but it's fine 👍

matthewmcgarvey (approved):

src/actions/guides/database/querying.cr
jwoertink: I originally did this in the one-liner, but there was so many `&` that it was kind of hard to read 😂

PR #701, opened 2021-07-01
Add note about using base validation. Slight re-org

paulcsmith:
Wanted to document how to do simple deletes. I think there are plenty of cases where you just want to do a quick delete, so I think this will be helpful to start with in the guides


stephendolan (approved):
Nice!

src/actions/guides/database/deleting_records.cr
```diff
@@ -13,33 +13,52 @@ class Guides::Database::DeletingRecords < GuideAction
...
+    ```crystal
+    server = ServerQuery.find(123)
+    Server::DeleteOperation.destroy!(server)
+    ```
+
+    If the record fails to be deleted, an `Avram::InvalidOperationError` will be raised.
+
+    ### Setting up a custom DeleteOperation
+
+    You can customize a DeleteOperations with callbacks and validations.  These
```
jwoertink:
```suggestion
    You can customize DeleteOperations with callbacks and validations.  These
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -13,33 +13,52 @@ class Guides::Database::DeletingRecords < GuideAction
...
-        DeleteServer.destroy(server) do |operation, deleted_server|
+        # The second block argument can be completely removed.
+        # It is shown here with a `_` to show that it can be used if needed.
+        DeleteServer.destroy(server) do |operation, _deleted_server|
           if operation.deleted?
             redirect to: Servers::Index
           else
             flash.danger = "Could not delete"
-            html Servers::EditPage, server: deleted_server
+            html Servers::EditPage, server: server
```
jwoertink: The only thing with this is if the record is soft deleted, the `server` wouldn't know that, but the `deleted_server` would.

jwoertink (commented):
Nice! Just left a few comments, but this definitely makes it a bit more clear 😄

src/actions/guides/database/deleting_records.cr
paulcsmith: Nice catch!

src/actions/guides/database/deleting_records.cr
paulcsmith: Ohhh, I didn't read about the new `soft_delete` changes. Will make those changes and re-ping for review later. Thanks! I would not have caught that :P

paulcsmith: Oh wait, can you do `operation.soft_deleted?` if so then that could work instead of using the `deleted_server`?

jwoertink: No. `operation.deleted?` will return `true` if the record was soft deleted though.

paulcsmith: Ah ok so you don't need the returned serve if you use `operation.deleted?` if so I'll document that. That way you don't need to use the second block arg in most (not all) cases.

jwoertink (approved):
Feel free to merge when ready!

jwoertink:
Gonna need a rebase here

src/actions/guides/database/deleting_records.cr
```diff
@@ -13,33 +13,52 @@ class Guides::Database::DeletingRecords < GuideAction
...
-        DeleteServer.delete(server) do |operation, deleted_server|
+        # The second block argument can be completely removed.
+        # It is shown here with a `_` to show that it can be used if needed.
+        DeleteServer.delete(server) do |operation, _deleted_server|
           if operation.deleted?
             redirect to: Servers::Index
           else
-            flash.failure = "Could not delete"
-            html Servers::EditPage, server: deleted_server
+            flash.danger = "Could not delete"
```
jwoertink:
```suggestion
            flash.failure = "Could not delete"
```

The valid methods are `success`, `failure`, and `info`.

src/actions/guides/database/deleting_records.cr
```diff
@@ -13,33 +13,52 @@ class Guides::Database::DeletingRecords < GuideAction
...
+
     ```crystal
     # src/actions/servers/delete.cr
     class Servers::Delete < BrowserAction
       delete "/servers/:server_id" do
         server = ServerQuery.find(server_id)
 
-        DeleteServer.delete(server) do |operation, deleted_server|
+        # The second block argument can be completely removed.
+        # It is shown here with a `_` to show that it can be used if needed.
```
jwoertink: Though, this is true, it doesn't fit in this case because in the `else` branch, the 2nd arg is used.

src/actions/guides/database/deleting_records.cr
```diff
@@ -13,33 +13,52 @@ class Guides::Database::DeletingRecords < GuideAction
...
+        DeleteServer.delete(server) do |operation, _deleted_server|
```
jwoertink:
```suggestion
        DeleteServer.delete(server) do |operation, deleted_server|
```

Since both variables are used.

src/actions/guides/database/deleting_records.cr
```diff
@@ -13,33 +13,52 @@ class Guides::Database::DeletingRecords < GuideAction
...
+    ### Simple deletes
 
-    These classes go in your `src/operations/` directory, and will inherit from `{ModelName}::DeleteOperation`.
+    If you just want to delete a record without any validations or callbacks, the simplest way is to use the generated `{ModelName}::DeleteOperation`
+
+    For example, if we have a `Server` model, Lucky will generate a `Server::DeleteOperation` that you can use:
+
+    ```crystal
+    server = ServerQuery.find(123)
+    Server::DeleteOperation.destroy!(server)
```
jwoertink:
```suggestion
    Server::DeleteOperation.delete!(server)
```

jwoertink (changes requested):
@stephendolan not sure if you were wrapping this up or not, but there's a few more changes that are still missing. Just since we've renamed the method and such since this was originally pushed up.

stephendolan:
@jwoertink Yeah, just trying to close this out to clear the website PR queue :)

src/actions/guides/database/deleting_records.cr
```diff
@@ -59,19 +74,17 @@ class Guides::Database::DeletingRecords < GuideAction
...
-    > Currently bulk deletes with DeleteOperation are not supported.
-
-    If you need to bulk delete a group of records based on a where query, you can use `delete` at the end of your query.
-    This returns the number of records deleted.
+    You can also use the `destroy!` method if you don't need validations and expect delete's to work every time:
 
     ```crystal
-    # DELETE FROM users WHERE banned_at IS NOT NULL
-    UserQuery.new.banned_at.is_not_nil.delete
+    DeleteServer.destroy!(server)
```
jwoertink:
```suggestion
    DeleteServer.delete!(server)
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -59,19 +74,17 @@ class Guides::Database::DeletingRecords < GuideAction
     end
     ```
 
-    ### Bulk delete
+    ### Destroy and raise if it fails
```
jwoertink:
```suggestion
    ### Delete and raise if it fails
```

src/actions/guides/database/deleting_records.cr
```diff
@@ -59,19 +74,17 @@ class Guides::Database::DeletingRecords < GuideAction
...
 
-    > Currently bulk deletes with DeleteOperation are not supported.
-
-    If you need to bulk delete a group of records based on a where query, you can use `delete` at the end of your query.
-    This returns the number of records deleted.
+    You can also use the `destroy!` method if you don't need validations and expect delete's to work every time:
```
jwoertink:
```suggestion
    You can also use the `delete!` method if you don't need validations and expect deletes to work every time:
```

jwoertink (changes requested):
Ok, I missed these last time, but I think that's all of them!

jwoertink (approved):
sweet :+1:  Thanks for doing that @stephendolan

PR #707, opened 2021-07-06
Update guides to use LuckyEnv

hovsater:
Updates the guides to make use of `LuckyEnv` instead of `Lucky::Env`. See https://github.com/luckyframework/lucky_cli/pull/655 for details.

### DIFF
--- a/src/actions/guides/getting-started/configuration.cr
+++ b/src/actions/guides/getting-started/configuration.cr
@@ -20,19 +20,23 @@ class Guides::GettingStarted::Configuration < GuideAction
 
     ## Get the current environment
 
-    `Lucky::Env.name` will return the currently running environment
+    `LuckyEnv.environment` will return the currently running environment
 
-    You can also conditionally check the environment with `Lucky::Env.test?`,
-    `Lucky::Env.development?`, or `Lucky::Env.production?`
+    You can also conditionally check the environment with `LuckyEnv.development?`,
+    `LuckyEnv.test?`, or `LuckyEnv.production?`
 
     ## Set the current environment
 
-    `LUCKY_ENV=environment_name` to set the environment. By default it is `development`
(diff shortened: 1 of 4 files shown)


stephendolan:
@kevinsjoberg Thanks so much for submitting this! If you can fix that one failing check, this looks good to merge!

hovsater:
@stephendolan oh, totally missed that. Thanks for reminding me. 🙂 I've amended the original commit.

jwoertink:
I didn't look it up, but does 0.27.2 generate Lucky apps with LuckyEnv built in? If it does, then we should be fine. If not, then we need to wait until 0.28 drops before merging this.

src/actions/guides/getting-started/configuration.cr
```diff
@@ -20,19 +20,23 @@ class Guides::GettingStarted::Configuration < GuideAction
...
-    `LUCKY_ENV=environment_name` to set the environment. By default it is `development`
+    `LUCKY_ENV=environment_name` to set the environment. By default it is `development`.
 
     ## Add custom environment
 
     If you need a custom environment like `staging`, for example, you can add this option in `config/env.cr`.
-    This will give you the helper method `Lucky::Env.staging?` for use in your app.
+    This will give you the helper method `LuckyEnv.staging?` for use in your app.
+
+    ```
```
jwoertink:
```suggestion
    ```crystal
```

jwoertink (changes requested):
I think just the syntax for highlighting consistency, and then we're good to merge! Thanks for doing this.

src/actions/guides/getting-started/configuration.cr
hovsater: Great suggestion. 👍

jwoertink (approved):
sweet!

PR #717, opened 2021-07-14
Document routing style

### LINKED ISSUE #703: Document underscore enforcer
https://github.com/luckyframework/lucky/pull/1536

This is handled automatically so not urgent that we document it. We also have nice compile-time errors that explain what to do. So this can go in a small note somewhere. Doesn't need a lot

hovsater:
Add documentation regarding [routing style checks](https://github.com/luckyframework/lucky/pull/1536/).

Screenshot 2021-07-14 at 21 41 59

Closes #703.

### DIFF
--- a/src/actions/guides/http_and_routing/routing_and_params.cr
+++ b/src/actions/guides/http_and_routing/routing_and_params.cr
@@ -247,6 +247,33 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
     end
     ```
 
+    ## Routing style
+
+    By default Lucky ensures that all routes adhere to the same style. All
+    route paths are expected to use underscores unless you opt-out or change
+    the style check. You can opt-out from style checking by including
+    `Lucky::SkipRouteStyleCheck` in your action.
+
+    ```crystal
+    # src/actions/users/show.cr
+    class Users::Show < BrowserAction
+      include Lucky::SkipRouteStyleCheck
+
+      get "/users/user-:some_user_id" do
+        plain_text "Requested user id: \#{some_user_id}"
+      end
+    end
+    ```
+
+    Or, skipping checking altogether by removing
+    `Lucky::EnforceUnderscoredRoute` from `src/actions/browser_action.cr`.
+
+
+    Similarly, a custom style check can be added by adding the
+    `enforce_route_style` macro in your action. Or, for all actions by adding
+    it to `src/actions/browser_action.cr`. See [Lucky::EnforceUnderscoredRoute](https://github.com/luckyframework/lucky/blob/master/src/lucky/enforce_underscored_route.cr)
+    for an example.
+
     ## Memoization
 
     As your application gets larger, you may need to write helper methods that run expensive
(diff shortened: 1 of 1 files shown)


src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -247,6 +247,33 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+    route paths are expected to use underscores unless you opt-out or change
+    the style check. You can opt-out from style checking by including
+    `Lucky::SkipRouteStyleCheck` in your action.
+
+    ```crystal
+    # src/actions/users/show.cr
+    class Users::Show < BrowserAction
+      include Lucky::SkipRouteStyleCheck
+
+      get "/users/user-:some_user_id" do
```
jwoertink: I'm not sure this is a valid route 🤔  I think the variable needs to be right after the `/` like `/users/:some_user_id` otherwise it would try to match a literal `site.com/users/user-:some_user_id`. Maybe this could be an action like `Guides::GettingStarted` with the route `"/guides/getting-started"`?

hovsater: @jwoertink you're right. Just checked the `PathPart` implementation. I'll update the example accordingly.

> Maybe this could be an action like Guides::GettingStarted with the route "/guides/getting-started"?

Or should we support defining variables wherever in the route path?

jwoertink: I don't think we should support defining variables wherever in the route path. If you wanted to prefix your values, you can define a `to_param` method on the model that does that when you pass the model to `Action.with`, I believe...

hovsater: Hmm, Rails supports this out of the box. Why wouldn't we allow specifying dynamic segments at arbitrary locations in the path?

jwoertink: > Rails supports this

that's true. I guess in Rails you can do stuff like `/thing(.:format)`, but that doesn't work in Lucky.

Personally, I'm not a fan of this because I think it's confusing to read, and similar to us restricting all routes to use `_`, Lucky tries to enforce a specific coding style to keep things consistent and thus less buggy. However, maybe this is something we can open a discussion on. I'll open up the issue to start the conversation to see what others think. Maybe the benefits of doing that outweigh the potential complications? 🤔

hovsater: Sounds fair to me. 🙂 Let's revert the example here, and continue the discussion in a separate issue. 👍

jwoertink:
```suggestion
    # src/actions/users/show.cr
    class Guides::GettingStarted < BrowserAction
      include Lucky::SkipRouteStyleCheck

      get "/guides/getting-started" do
```

src/actions/guides/http_and_routing/routing_and_params.cr
```diff
@@ -247,6 +247,33 @@ class Guides::HttpAndRouting::RoutingAndParams < GuideAction
...
+        plain_text "Requested user id: \#{some_user_id}"
```
jwoertink:
```suggestion
        plain_text "Get started"
```

jwoertink (changes requested):
Just need this change, and it's good 👍

jwoertink (approved):
Alright, I was able to apply those suggestions. I'm gonna merge this in. Thanks for handling this!

PR #719, opened 2021-07-15
Replace myhtml with lexbor

hovsater:
Another attempt since #365 at switching [myhtml](https://github.com/kostya/myhtml/) to [lexbor](https://github.com/kostya/lexbor). Tests are passing, but I guess some manual validation would be nice as well.

I hope to address #358 in another pull request.


jwoertink:
I'm cool with this; however, the issue was never really the code. It was deploying to Heroku where the show stopper was. I think we had to change the buildpack to include cmake or something like that 🤔  This might be a question for @stephendolan  but is there a way to handle that without editing the existing buildpack?

hovsater:
@jwoertink I believe `cmake` is already required for the `cr-cmark-gfm` shard? I'd be happy looking into the Heroku-specifics as well if it turns out to be a problem. 🙂

The benefit of making the switch to `lexbor` in any case is relying on maintained software.

jwoertink (approved):
Yeah, I think we have to switch anyway... I think we can just merge it in, deploy, and see where it tanks... Then open up an issue to get that fixed... Not the best workflow, but only PRs from me will do test deployments, so there's not an easy way to really see the errors 😅

PR #721, opened 2021-07-17
Bump Crystal to 1.1.0

hovsater:
Bumps Crystal to version `1.1.0`.


hovsater:
The check first failed due to a compiler not being present. I added `build-essential` to the workflow to fix that, but apparently the checks ran without this change. Not sure why.

jwoertink:
Thanks for giving this a shot. I think we can not worry about this for now. I will take care of updating to 1.1.0 after Lucky supports it in the next release. I have been changing up how the CI works in other repos anyway 😃

hovsater:
@jwoertink sounds good. I'll just close this then. 🙂

PR #724, opened 2021-07-17
Update the links in the HTTP status code list.

takanotume24:
HTTP status codes could not be built with the notation of this link. The notation was changed to quote the current Crystal's HTTP library.


jwoertink (approved):
Great catch! I forgot we used to have our own statuses 😂 Thanks for submitting this PR

PR #729, opened 2021-07-21
Use failure instead of danger

ghost:
Change the failed flash message from `danger` to `failure`


matthewmcgarvey (approved):
Thanks for catching that 👍

jwoertink (approved):
Thanks!

PR #756, opened 2021-07-30
adding guides for raw SQL with args.

### LINKED ISSUE #750: Document custom raw SQL with args
https://luckyframework.org/guides/database/raw-sql

If you need to pass in arguments to your custom SQL, it gets a little tricky. Here's an example of passing in a UUID

```crystal
class Report
  DB.mapping(id: UUID, type: String)

  def self.find(id : UUID)
    query = <<-SQL
    SELECT id, type FROM reports WHERE id = $1
    SQL

    AppDatabase.run do |db|
      db.query_all query, args: [id], as: Report
    end
  end
end
```

Note that this is using `$1`. If we had a second arg, we'd have to set `$2` and so on. This is in contrast to the placeholders we use in normal queries like `ReportQuery.new.where("id = ?", id)`

jwoertink:
Fixes #750

### DIFF
--- a/src/actions/guides/database/raw-sql.cr
+++ b/src/actions/guides/database/raw-sql.cr
@@ -22,7 +22,8 @@ class Guides::Database::RawSql < GuideAction
 
     ## Custom Queries
 
-    Since we have direct access to the database instance, we can run whatever query we want as long as it's valid sql for the [postgresql driver that Lucky uses](https://github.com/will/crystal-pg).
+    Since we have direct access to the database instance, we can run whatever query we want as
+    long as it's valid sql for the [postgresql driver that Lucky uses](https://github.com/will/crystal-pg).
 
     Let's create a query with a non-trivial `SELECT` statement that we can map to a Crystal class.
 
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/raw-sql.cr
```diff
@@ -46,15 +47,17 @@ class Guides::Database::RawSql < GuideAction
...
+    it simple to map a database query to a class by defining the keys and types of each column.
 
     Let's create a `ComplexPost` class in our models folder and define the database mapping.
 
     ```crystal
     # src/models/complex_post.cr
     class ComplexPost
       DB.mapping({
-        id: Int32,
+        id: Int64,
```
jwoertink, line 60: because this is the more likely case now.

src/actions/guides/database/raw-sql.cr
```diff
@@ -22,7 +22,8 @@ class Guides::Database::RawSql < GuideAction
 
     ## Custom Queries
 
-    Since we have direct access to the database instance, we can run whatever query we want as long as it's valid sql for the [postgresql driver that Lucky uses](https://github.com/will/crystal-pg).
+    Since we have direct access to the database instance, we can run whatever query we want as
```
jwoertink, line 25: same wording, just formatting

matthewmcgarvey (approved):

PR #802, opened 2021-09-26
Make #raw text helper more searchable

stephendolan:
Searching "raw" doesn't return the section that documents `raw`. I'm hoping that adding it in parentheses will help 😄


jwoertink (approved):
:+1:

PR #877, opened 2021-12-22
New links and tagging system on Awesome list

### LINKED ISSUE #875: More for the Awesome list
Shards:
* https://github.com/GrottoPress/lucille
* https://github.com/GrottoPress/bill

Sites:
* https://dendrorithms.com/
* https://reviewmy.email/

jwoertink:
Fixes #875

With the tagging system, we can now tag links as "Adult". This doesn't mean it's "porn", but more that the link *may* contain content not suitable for younger people. When clicking on a link tagged as "Adult", you get a confirmation before being taken to that URL.

Also note that we can add multiple tags to a single link.

Screen Shot 2021-12-22 at 2 48 32 PM

Screen Shot 2021-12-22 at 2 48 38 PM

### DIFF
--- a/src/components/awesome_link.cr
+++ b/src/components/awesome_link.cr
@@ -1,7 +1,7 @@
 class AwesomeLink < BaseComponent
   enum Tags
     Crypto
-    Adult # this link may contain content not suitable for persons under 18 years old
+    Adult # this link may contain content that is age restricted
   end
 
   needs text : String
@@ -33,9 +33,9 @@ class AwesomeLink < BaseComponent
   private def render_tags
     (tags || [] of Tags).each do |tag|
       if tag.adult?
-        span "18+",
+        span "Adult",
           class: "bg-red-500 text-white py-1 px-3 rounded text-xs ml-2",
-          title: "This link may contain content not suitable for persons under 18 years of age"
+          title: "This site may contain sensitive or age restricted content. Discretion is advised."
       else
         span tag.to_s,
           class: "bg-lucky-dark-green text-white py-1 px-3 rounded text-xs ml-2"
--- a/src/components/awesome_link.cr
+++ b/src/components/awesome_link.cr
@@ -1,14 +1,21 @@
 class AwesomeLink < BaseComponent
+  enum Tags
+    Crypto
+    Adult # this link may contain content not suitable for persons under 18 years old
+  end
+
   needs text : String
   needs url : String
   needs description : String?
+  needs tags : Array(Tags)?
 
   def render
     li do
-      a class: "block hover:bg-gray-50", href: url, target: "_blank" do
+      a class: "block hover:bg-gray-50", href: url, target: "_blank", data_confirm: confirmation_text do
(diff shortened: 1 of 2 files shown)


src/pages/learn/awesome_lucky/index_page.cr
```diff
@@ -94,6 +94,16 @@ class Learn::AwesomeLucky::IndexPage < PageLayout
         text: "Scribe",
         url: "https://scribe.rip/",
         description: "Alternative frontend to Medium.com"
+      mount AwesomeLink,
+        text: "DendroRithms",
+        url: "https://dendrorithms.com/",
+        description: "DendroRithms is a generative art installation inspired by tree growth rings",
+        tags: [AwesomeLink::Tags::Crypto]
```
jwoertink: @wout is this a fair tag to use?

wout: Ah, you got there before me. :slightly_smiling_face: 
Yes, I think this is looking great!
Thanks for doing this.

matthewmcgarvey (approved):

src/components/awesome_link.cr
```diff
@@ -1,14 +1,21 @@
 class AwesomeLink < BaseComponent
+  enum Tags
+    Crypto
```
edwardloveall, line 3: Seems strange to me that we only have two tags. I get why we're tagging Adult, but does Crypto need to be treated specially also? If not should we also have tags for everything else?

src/components/awesome_link.cr
```diff
@@ -22,4 +29,25 @@ class AwesomeLink < BaseComponent
       end
     end
   end
+
+  private def render_tags
+    (tags || [] of Tags).each do |tag|
+      if tag.adult?
+        span "18+",
```
edwardloveall: How about `Adult` instead of `18+`? Age leans more into the legal aspect (in particular) of pornography (in particular), in the united states (in particular). Even then the age varies state to state. To address a world-wide audience and themes other than pornography, I think `Adult` could be both more correct and help inform the reader so they can make the best decision for themselves.

What concept do you want to communicate with this text?

src/components/awesome_link.cr
jwoertink: I added the `Crypto` only for the purposes of being able to test a multiple tagging system. I actually had an idea for tags in the shards section where shards that are made specifically for Lucky could be tagged as such. Then there's other shards that work with Lucky, but they're not necessarily "lucky" shards (e.g. multi_auth, sitemapper, etc...). I could drop it for now, and then we can talk about what would make most sense in a separate PR.

src/components/awesome_link.cr
jwoertink: That's true... and actually, that might make more sense anyway. I did a "beer and wine" online store a few years back, and that would be one that would fall in to this category (as well as sites like dispensaries or whatever...). Alcohol related sites generally need a 21+ anyway (at least in the U.S.). Good call.

matthewmcgarvey (approved):

PR #929, opened 2022-02-17
Add Dependabot Automerge Github action

grepsedawk:
@jwoertink mentioned there aren't JS tests so this might not be the best move, but this merges minor/patch dependabot PRs automatically

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,14 @@
+name: Auto-merge minor/patch
+on:
+  schedule:
+    - cron: "0 * * * *"
+jobs:
+  test:
+    name: Auto-merge minor and patch updates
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,14 @@
...
+    - cron: "0 * * * *"
+jobs:
+  test:
+    name: Auto-merge minor and patch updates
+    runs-on: ubuntu-latest
+    steps:
+      - uses: koj-co/dependabot-pr-action@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          merge-minor: true
```
grepsedawk:
```suggestion
          merge-minor: false
```
Alternatively, don't automerge minors

PR #949, opened 2022-03-08
Change from turbolinks to turbo

grepsedawk:
Reverts the previous revert of turbolinks -> turbo
Fixes html conversions page, which resulted in a turbo error previously

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -1,13 +1,9 @@
 /* eslint no-console:0 */
 
 // RailsUjs is *required* for links in Lucky that use DELETE, POST and PUT.
-import RailsUjs from "rails-ujs";
+require("@rails/ujs").start();
+require("@hotwired/turbo");
 
-// Turbolinks is optional. Learn more: https://github.com/turbolinks/turbolinks/
-import Turbolinks from "turbolinks";
-
-RailsUjs.start();
-Turbolinks.start();
 
 import hljs from 'highlight.js/lib/core';
 
@@ -43,7 +39,7 @@ hljs.registerLanguage('diff', diff);
 
 import docsearch from '@docsearch/js';
 
-document.addEventListener("turbolinks:load", function () {
+document.addEventListener("turbo:load", function () {
 
   docsearch({
     container: '#docsearch',
--- a/src/pages/html_conversions/new_page.cr
+++ b/src/pages/html_conversions/new_page.cr
@@ -7,26 +7,27 @@ class HtmlConversions::NewPage < PageLayout
       h2 title, class: "text-3xl mb-4"
 
       para "Lucky uses Crystal classes and methods to generate HTML. It may sound crazy at first, but the advantages are numerous. Never accidentally print nil to the page, extract and share partials using regular methods. Easily read an entire page by looking at just the render method. Text is automatically escaped for security. And it’s all type safe. That means no more unmatched closing tags, and never rendering a page with missing data."
-
-      form_for HtmlConversions::Create, class: "my-4 space-y-2" do
-        div class: "space-y-1" do
-          h3 "HTML Input", class: "text-2xl"
-          textarea input, name: "input", class: "appearance-none mt-2 border-2 border-gray-200 rounded w-full py-2 px-4 leading-tight focus:outline-none focus:bg-white", placeholder: "Paste your HTML here", attrs: [:required], rows: 4
-        end
-
-        render_down_arrow
-        render_convert_button
-        render_down_arrow
-
-        div class: "space-y-1" do
-          div class: "flex items-center justify-between" do
-            h3 "Lucky Output", class: "text-2xl"
-            render_copy_button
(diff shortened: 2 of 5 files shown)


grepsedawk:
This is the line that differentiates now from before: https://github.com/luckyframework/website/pull/949/files#diff-5cad4259e0176077512aa1c6ed2783c520d3295556400d76cbcd6726445d576bR10

src/js/app.js
```diff
@@ -43,7 +39,10 @@ hljs.registerLanguage('diff', diff);
 
 import docsearch from '@docsearch/js';
 
-document.addEventListener("turbolinks:load", function () {
+document.addEventListener("turbo:load", function () {
+  document.querySelectorAll('pre code').forEach((block) => {
+    hljs.highlightElement(block);
+  });
```
grepsedawk:
```suggestion
```

grepsedawk:
Another thing I added while I was in there was auto-submitting form:

https://user-images.githubusercontent.com/6721355/157188185-fbd5a8b3-271f-4865-ac76-6a960c28d13b.mov

It makes it feel nice and live while using it, without the need for a convert button

Includes a 1 second debounce so we're not bombarding the server.

src/js/app.js
```diff
@@ -55,4 +51,14 @@ document.addEventListener("turbolinks:load", function () {
...
+
+  let autosaveKeyupTimer
+  document.addEventListener('keyup', (event) => {
+    if (event.target.form.classList.contains('autosave')) {
+      clearTimeout(autosaveKeyupTimer)
+      autosaveKeyupTimer = setTimeout(() => {
+        event.target.form.requestSubmit()
+      }, 1000)
+    }
+  })
```
jwoertink: What is this needed for?

src/pages/html_conversions/new_page.cr
```diff
@@ -7,17 +7,16 @@ class HtmlConversions::NewPage < PageLayout
...
           textarea input, name: "input", class: "appearance-none mt-2 border-2 border-gray-200 rounded w-full py-2 px-4 leading-tight focus:outline-none focus:bg-white", placeholder: "Paste your HTML here", attrs: [:required], rows: 4
         end
+      end
 
-        render_down_arrow
-        render_convert_button
-        render_down_arrow
+      render_down_arrow
 
+      tag "turbo-frame", id: "html-conversion" do
```
jwoertink, line 10: Is this required for all forms with turbo?

grepsedawk: No, only forms that redirect/render to themselves

src/js/app.js
grepsedawk: https://github.com/luckyframework/website/pull/949#issuecomment-1061488005
I can drop this diff, but it was an easy add with turbo involved

src/pages/html_conversions/new_page.cr
grepsedawk: It's just a catch for turbo to say "hey I don't know if you want me to morph the page or reload the page"
So you can either add a "target=_top" to the form to force a normal form submit, or you can add a frame to tell it what part of the page to update

src/js/app.js
jwoertink: It's probably ok. I'm just trying to understand what is required for turbo to work out of the box. We're using this as a testing ground before adding it in to Lucky core. So I don't want to confuse what's absolutely needed vs what's sugar.

grepsedawk: Ignore the secondary commit for the fluff
The only thing required for the turbo form to post to itself is the frame so it knows what to update

src/pages/html_conversions/new_page.cr
jwoertink: Ah, ok. So in this case, it would be up to the developer to just "know" that this had to be added? :thinking:  I'm just trying to think from a Lucky standpoint, can we catch this sort of error in development, or would it be a case of just adding documentation around it....

grepsedawk: https://github.com/hotwired/turbo-rails/issues/12#issuecomment-754629885
I think it's in incredibly rare to re-render the same page with a 200 http code. However, even once people do go outside of this box, there is a turbo error already thrown.

grepsedawk:
Is there anything I can do to make this easier to review, too? I do realize this is kind of a mess with the full turbo install and the changes -- if it would be helpful I could split up the "revert" and "fix form" parts 😄

stephendolan:
I do actually think it would be cleaner to just stick with turbolinks -> turbo in this PR, and separately refactor the routes, set up the form autosubmit, etc.

The large scope was what's kept me from hitting the "Approve" button so far.

grepsedawk:
Pushed just the turbo changeset now

stephendolan (approved):
As long as the page is working, this seems fine until we can provide cleaner default support.

grepsedawk:
(force push was just to resolve conflicts from dependabot merges)

PR #951, opened 2022-03-10
Rename html_conversions page from index to new

grepsedawk:
The class inside references new, so for people using filename based
fuzzy searching, this makes it marginally easier to get to this file


robcole:
> The class inside references new, so for people using filename based
> fuzzy searching, this makes it marginally easier to get to this file

This would have saved me a couple of minutes at the very least when digging into the turbo stuff. 👍

matthewmcgarvey (approved):

PR #1005, opened 2022-05-31
Update i18n instruction for Rosetta 0.7.0

wout:
I've just released Rosetta v0.7.0 which includes a number of [improvements for Lucky](https://github.com/wout/rosetta/blob/main/src/rosetta/lucky/integration.cr). Integration is now better for:
- `Lucky::FlashStore`
- `Lucky::SpecialtyTags`
- `Lucky::FormHelpers`

It also adds a `Rosetta::AvramBackend` for validation translations, and there's a special [initializer flag for Lucky](https://wout.github.io/rosetta/v0.7.0/getting_started/#using-lucky).

Finally, the configuration API has changed a bit. Configuring the default locale, available locales and pluralization rules is now done using annotations instead of constants.

Also, in recent benchmarks, Rosetta is now 12 to 700 times faster than similar libraries. :nerd_face:

### DIFF
--- a/src/actions/guides/frontend/internationalization.cr
+++ b/src/actions/guides/frontend/internationalization.cr
@@ -67,7 +67,7 @@ class Guides::Frontend::Internationalization < GuideAction
     bin/rosetta --init --lucky
     ```
 
-    This command will geneate:
+    This command will generate:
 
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
Awesome! Thanks for the update :smile:

PR #1022, opened 2022-06-05
Adding in docs for the previous_url method. 

### LINKED ISSUE #964: Add docs on previous_url method
https://github.com/luckyframework/lucky/pull/1641

jwoertink:
Fixes #964

### DIFF
--- a/src/actions/guides/frontend/rendering_html.cr
+++ b/src/actions/guides/frontend/rendering_html.cr
@@ -450,6 +450,25 @@ class Guides::Frontend::RenderingHtml < GuideAction
     # => true
     ```
 
+    ### Previous URL
+
+    Lucky also has a `previous_url` helper method you can use on your pages.
+
+    It returns the url of the page that issued the request (the referrer)
+    if possible, otherwise redirects to the provided default fallback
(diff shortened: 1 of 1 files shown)


src/actions/guides/frontend/rendering_html.cr
```diff
@@ -450,6 +450,25 @@ class Guides::Frontend::RenderingHtml < GuideAction
     # => true
     ```
 
+    ### Previous URL
+
+    Lucky also has a `previous_url` helper method you can use on your pages.
+
+    It returns the url of the page that issued the request (the referrer)
+    if possible, otherwise redirects to the provided default fallback
```
matthewmcgarvey:
```suggestion
    if possible, otherwise it uses the provided default fallback
```

Since I confused this with `redirect_back`, I figured it best to remove "redirect" language

matthewmcgarvey (approved):

src/actions/guides/frontend/rendering_html.cr
jwoertink: That makes sense. Thanks!

PR #1044, opened 2022-07-24
Update installation instructions for M1 & Intel Macs

grvm:
By default the Homebrew directory is /opt/homebrew on M1 Macs & /usr/local for Intel Macs

### DIFF
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -173,16 +173,11 @@ class Guides::GettingStarted::Installing < GuideAction
     ### 6. Move the generated binary to your path
 
     This will let you use `lucky` from the command line.
-
-    macOS (Intel)
+    
     ```plain
     cp bin/lucky /usr/local/bin
     ```
-    
-    macOS (M1)
-    ```plain
-    cp bin/lucky /opt/homebrew/bin
-    ```
+    Or anywhere else you deem fit
 
     ### 7. Check installation
 
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -173,7 +173,6 @@ class Guides::GettingStarted::Installing < GuideAction
     ### 6. Move the generated binary to your path
 
     This will let you use `lucky` from the command line.
-    
     ```plain
     cp bin/lucky /usr/local/bin
     ```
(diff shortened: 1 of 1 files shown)


src/actions/guides/getting-started/installing.cr
```diff
@@ -162,9 +174,15 @@ class Guides::GettingStarted::Installing < GuideAction
...
     This will let you use `lucky` from the command line.
 
+    macOS (Intel)
     ```plain
     cp bin/lucky /usr/local/bin
     ```
+    
+    macOS (M1)
+    ```plain
+    cp bin/lucky /opt/homebrew/bin
```
jwoertink: This section isn't the homebrew installation. This is when you build the lucky binary by hand. I believe it should still be put in to `/usr/local/bin` even on M1, right?

grvm: Hey @jwoertink On M1 macs, `/usr/local` is owned by root, not the regular user, so this command errors out with permission denied

jwoertink: We were having a little discussion on this over in the Lucky discord if you're interested https://luckyframework.org/chat

I think my main hesitation here is that if you're not using homebrew, and you want to install Lucky manually, you wouldn't want to copy to `/opt/homebrew/bin` because then you'd have to manually create that directory and add it to $PATH first. However, I also think that someone not using homebrew would probably be the edge case anyway, so maybe it's fine?... 

I'm also thinking, maybe we just say `sudo cp bin/lucky /usr/local/bin`. I looked at mine on Pop_OS, and it's owned by `root` except for the `lucky ` binary which is owned by me. @robacarp made the suggestion that maybe that step is outside of the scope here? We could  maybe change it to just say "Move the `bin/lucky` binary to a directory in your $PATH so it's accessible anywhere.". Something along those lines...

What do you think @grvm ?

grvm: Yeah, that makes sense. I'll revert it to the original command with sudo & add a line "or anywhere in your path"

src/actions/guides/getting-started/installing.cr
```diff
@@ -161,10 +173,11 @@ class Guides::GettingStarted::Installing < GuideAction
     ### 6. Move the generated binary to your path
 
     This will let you use `lucky` from the command line.
-
+
```
```suggestion

```

I think this is a linting error. Gonna update this to kick off CI.

jwoertink:
Guess not :sweat_smile:  I'll just merge and fix the lint after. Thanks for doing this @grvm

grvm:
Hey, Thanks for merging @jwoertink sorry this got lost in notifications. I can't see the ilnt logs now, let me know if I can be of any help :)

jwoertink:
Thanks! I think it's all good now :smile:

PR #1084, opened 2022-09-30
Added blog post for 1.0.0-rc1

jwoertink:


src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    [Turbolinks](https://github.com/turbolinks/turbolinks-classic).
+    This library helped to make pages feel even faster, and who doesn't love speed?
+    Lucky included this library by default, but it has since been deprecated in favor
+    of [Turbo](https://turbo.hotwired.dev/). Although Lucky doesn't include Turbo currently,
+    we would still like to encourage users to give it a shot, and let us know if this
+    is something you'd like to see included by default.
+
+    Along those same lines, Lucky also came out-of-the-box with [Laravel-Mix](https://laravel-mix.com/)
+    which is "An elegant wrapper around Webpack". This made bringing javascript to Lucky
+    seemless, and easy. However, the Laravel framework has [moved to Vite](https://laravel-news.com/vite-is-the-default-frontend-asset-bundler-for-laravel-applications)
```
matthewmcgarvey:
```suggestion
    seamless, and easy. However, the Laravel framework has [moved to Vite](https://laravel-news.com/vite-is-the-default-frontend-asset-bundler-for-laravel-applications)
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    as their default front-end. Laravel-Mix may or may not stick around, but with the
+    uncertianty, the 1.0.0-rc1 release opens up the possibility to use [Vite](https://vitejs.dev/)
```
matthewmcgarvey:
```suggestion
    uncertainty, the 1.0.0-rc1 release opens up the possibility to use [Vite](https://vitejs.dev/)
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    while keeping with the type-safe traditions Lucky brings.
+
+    So why even include a javascript build system with Lucky?
+
+    One of Lucky's goals is to be type-safe and help catch bugs in development. This
+    is done many ways, but one of the most important ways is when it comes to loading
```
matthewmcgarvey:
```suggestion
    is done in many ways, but one of the most important ways is when it comes to loading
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
+class Lucky100rc1Release < BasePost
+  title "Lucky reaches the 1.0 ascent with the first release candidate"
```
paulcsmith: What do you think about this:

```suggestion
  title "Lucky begins the 1.0 ascent with the first release candidate"
```

I think that might flow better than "reaches the ascent"

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    Lucky 1.0.0-rc1 has been released, and it's a huge milestone
+    for the framework and community.
+    TEXT
+  end
+
+  def content : String
+    <<-MD
+    As many developers know, reaching a 1.0 milestone is huge, and something to
+    be celebrated. We're stoked to release the first prelease of the 1.0 series with
+    `rc1`. We will be using this and the next few release candidates to clean up
```
paulcsmith:
```suggestion
    As many developers know, reaching a 1.0 milestone is huge, and something to
    be celebrated. We're excited to publish the first release candidate of Lucky 1.0. 
    We will be using this and the next few release candidates to clean up
```

Just thinking "stoked" might not be a word most non-English speakers know and may not be easily looked up. So maybe "excited"? Up to you though!

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    the rough edges of the framework, and make sure all of the dependencies in the
+    Lucky ecosystem are all solid and ready to take us to the next level.
```
paulcsmith: Since "all of the dependencies" was mentioned

```suggestion
    Lucky ecosystem are solid and ready to take us to the next level.
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    Many sites (including this one) have no need for a databaase. In some cases,
+    your data comes from a 3rd party API, or maybe you need a custom database
+    engine other than PostgreSQL.
+
+    In this release, Avram is no longer a dependency of Lucky, but instead,
+    Lucky is now an optional dependency of Avram. If you're using Avram outside
+    of Lucky, then nothing changes for you. However, for Lucky users using Avram,
+    you'll now have to include it as an additional dependency. Be sure to add it
+    to your `shard.yml` file, and require it in your `src/shards.cr`.
```
paulcsmith: Does this mean that by default Avram is not included anymore when you use `lucky init`?  Or is this saying that for _existing_ Lucky websites you need to explicitly add it?

If the latter then maybe:

```suggestion
    of Lucky, then nothing changes for you. However, for existing Lucky users upgrading to `rc1` that need Avram,
    you'll now have to include it as an additional dependency. Be sure to add it
    to your `shard.yml` file, and require it in your `src/shards.cr`.
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    # after this release
+    SavePost.create(params) do |o, p|
+      # ...
+    end
+    ```
+
+    This ended up being quite a big breaking change. The original Params implementation
+    assumed the type was `Hash(String, String)`, but this meant you couldn't have an
+    array value. Now the type assumes `Hash(String, Array(String))` which is actually
```
paulcsmith, line 75: May me good to share how these params are accessed or what the type changes means for users. Or a link to docs or code so people can learn more

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    # post:tags[]=lucky&post:tags[]=crystal
+    tags = params.nested_arrays("post")["tags"]
+    SavePost.create(params, tags: tags) do |o, p|
+      # ...
+    end
+
+    # after this release
+    SavePost.create(params) do |o, p|
+      # ...
+    end
```
paulcsmith: Super nitpicky but maybe naming these variables will be clearer to new Lucky users?

```suggestion
    SavePost.create(params, tags: tags) do |operation, post|
      # ...
    end

    # after this release
    SavePost.create(params) do |operation, post|
      # ...
    end
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    As we all know, the javascript ecosystem changes in leaps and bounds daily.
+    This makes it very difficult to keep up.
+
+    In 2017, when Lucky began, the Rails core team had a neat library called
+    [Turbolinks](https://github.com/turbolinks/turbolinks-classic).
+    This library helped to make pages feel even faster, and who doesn't love speed?
+    Lucky included this library by default, but it has since been deprecated in favor
+    of [Turbo](https://turbo.hotwired.dev/). Although Lucky doesn't include Turbo currently,
+    we would still like to encourage users to give it a shot, and let us know if this
+    is something you'd like to see included by default.
```
paulcsmith: I am not sure from reading this what has changed. 

> but it has since been deprecated in favor of [Turbo]

Does that mean we have deprecated Turbolinks in favor of Turbo, or that Rails has and we have deprecated both? And when it says encourage users to give it a shot do you mean encourage people to add Turbo and see what they think?

If so then maybe something like this:

```suggestion
    [Turbolinks](https://github.com/turbolinks/turbolinks-classic).
    This library helped to make pages feel even faster, and who doesn't love speed?
    Lucky included Turbolinks by default, but Rails has since deprecated it in favor
    of [Turbo](https://turbo.hotwired.dev/). Lucky now doesn't include either library, but
    we would still like to encourage users to give Turbo a shot, and let us know if it
    is something you'd like to see included by default.
```

I'd also be curous to hear why we removed Turbolinks without replacing it with Turbo. I'm not saying it was bad/good just curious why. I haven't been following it too closely so am not sure of the pros/cons and I think readers may also be curious about why we removed it

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    Along those same lines, Lucky also came out-of-the-box with [Laravel-Mix](https://laravel-mix.com/)
+    which is "An elegant wrapper around Webpack". This made bringing javascript to Lucky
+    seemless, and easy. However, the Laravel framework has [moved to Vite](https://laravel-news.com/vite-is-the-default-frontend-asset-bundler-for-laravel-applications)
+    as their default front-end. Laravel-Mix may or may not stick around, but with the
+    uncertianty, the 1.0.0-rc1 release opens up the possibility to use [Vite](https://vitejs.dev/)
+    while keeping with the type-safe traditions Lucky brings.
```
paulcsmith, line 96: I think this line could probably be removed since it is explained later on. Up to you though :D

src/posts/lucky_1.0.0-rc1.cr
paulcsmith:
```suggestion
    One of Lucky's goals is to be type-safe and help catch bugs in development. One of the most important ways is when it comes to loading
```

I'd maybe cut the sentence down a bit but up to you :)

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    So why even include a javascript build system with Lucky?
+
+    One of Lucky's goals is to be type-safe and help catch bugs in development. This
+    is done many ways, but one of the most important ways is when it comes to loading
+    your assets in production. By utilizing a generated manifest file, Crystal can
```
paulcsmith: It generates Crystal code, but I think this is a Lucky specific feature

```suggestion
    your assets in production. By utilizing a generated manifest file, Lucky can
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    generate an internal Hash at compile-time that can be referenced in the app.
```
paulcsmith: I _believe_ it is a NamedTuple since Hash does not have compile-time safety for keys

```suggestion
    generate an internal `NamedTuple` at compile-time that can be referenced in the app.
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    This means in development, you can do `img src: asset("images/logo.svg")`, and
+    Lucky knows this file is located at `public/assets/images/logo.svg`. In production,
+    you may need to use a cache buster, or even a CDN where the filename may change to
+    say `assets.mysite.com/images/logo-abc123xyz.svg`. If this file is missing, then
```
paulcsmith: > In production, you may need to use a cache buster, or even a CDN where the filename may change to say `assets.mysite.com/images/logo-abc123xyz.svg`

I am not sure if this is saying that Lucky generates a cache busted version by appending a hash to the filename or if the user has to do that. I think we used to generate a unique identifier that was appended to the filename so it was cache busted automatically. Is that still the case or do people need to do that themselves? I think that may need to be clarified here but I'm not sure what Vite does

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    So why even include a javascript build system with Lucky?
+
+    One of Lucky's goals is to be type-safe and help catch bugs in development. This
+    is done many ways, but one of the most important ways is when it comes to loading
+    your assets in production. By utilizing a generated manifest file, Crystal can
+    generate an internal Hash at compile-time that can be referenced in the app.
+
+    This means in development, you can do `img src: asset("images/logo.svg")`, and
+    Lucky knows this file is located at `public/assets/images/logo.svg`. In production,
```
paulcsmith: May be
```suggestion
    This means in development, you can do `img src: asset("images/logo.svg")`, and
    Lucky knows this file is located at `public/assets/images/logo.svg`, and will raise an error at compile-time if that asset does not exist/
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    you may need to use a cache buster, or even a CDN where the filename may change to
+    say `assets.mysite.com/images/logo-abc123xyz.svg`. If this file is missing, then
+    your app will not compile. That means that last second production deploy you're
```
paulcsmith: I'd probably move the "if the file is missing" to the first sentence since it is disconnected by the cache busting sentence and easy to miss. What do you think?

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    doing on a Friday before a 3-day weekend won't have you back in the "office"
+    doing another deploy because the file isn't showing.
```
paulcsmith: Woo hoo!

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    ### Email layouts and callbacks
+
+    Carbon 0.3.0 has been released with a few neat updates. The first being
+    the ability to set email layouts. Just like your standard views, you may
+    have a very similar wrapper on your emails like contact footer info, or
+    your logo.
```
paulcsmith, line 122: Awesome! That is super helpful!

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    ```crystal
+    # src/emails/welcome_email.cr
+    class WelcomeEmail < BaseEmail
+      def initialize(@recipient : Carbon::Emailable)
+      end
+
+      to @recipient
+      from "info@myapp.com"
+      subject "Welcome to the site!"
+      layout my_email_layout
```
paulcsmith, line 136: This is more a Carbon suggestion than a doc/blog one, but I would have expected this to be the name of the file and not the folder. May be worth changing this so the default is always `src/emails/templates/layouts/{file}` and the `layout` is the name of the file, not the folder. I think that is more common in most languages and that way you have one folder with all your layouts rather than a folder for each kind of layout

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+      templates html, text
+    end
+    ```
+
+    Then in your templates, you can add `my_email_layout/layout.ecr`,
+    and use the `content` method to yield your email body
+
+    ```
+    # src/emails/templates/my_email_layout/layout.ecr
```
paulcsmith, line 145: I'm not sure, but should this have the content type in the filename?

```suggestion
    # src/emails/templates/my_email_layout/layout.html.ecr
```

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+      def initialize(@recipient : Carbon::Emailable, @friend : User)
+      end
+
+      before_send do
+        if !@recipient.wants_email_from_friend?(@friend)
+          self.deliverable = false
+        end
+      end
+
+      # the response is from the email service (i.e. sendgrid, etc...)
```
paulcsmith, line 173: Are there docs we could link to for this? I think this may be hard to understand without some further docs.

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    ```crystal
+    class FriendIsLiveEmail < BaseEmail
+      def initialize(@recipient : Carbon::Emailable, @friend : User)
+      end
+
+      before_send do
+        if !@recipient.wants_email_from_friend?(@friend)
+          self.deliverable = false
+        end
+      end
```
paulcsmith, line 171: This is pretty cool! Can you stack `before_send` calls like you can with `before_save` in operations? Like, could I have a `before_send` in `BaseEmail` and also one in inheriting classes and it calls both?

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    end
+    ```
+
+    ### More ORM updates
+
+    Working towards better compatibility with [CockroachDB](https://www.cockroachlabs.com/product/)
+    enums now support both `Int64` values, as well as the `@[Flags]` annotations making bitwise
+    permissions super easy to achieve in Lucky!
+
+    We've also added support for the `bytea` column with Crystal's `Byte` (`Slice(UInt8)`) alias.
```
paulcsmith: This is cool but I think the other stuff is even cooler. Maybe put the compatability and bye stuff lower down and the new `extract` method higher?

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+    Another amazing update is the addition of the `extract()` query method for timestamp columns.
+    This allows you to query for date parts of a `Time` object. For example, query for
+    all users that signed up in July.
+
+    ```crystal
+    UserQuery.new.created_at.extract(:month).eq(7)
+    ```
```
paulcsmith, line 192: This is freakin cool

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+
+  def published_at : Time
+    Time.utc(year: 2022, month: 10, day: 1)
+  end
+
+  def summary : String
+    <<-TEXT
+    Lucky 1.0.0-rc1 has been released, and it's a huge milestone
+    for the framework and community.
+    TEXT
```
paulcsmith, line 13: 🥳

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+  end
+
+  def content : String
+    <<-MD
+    As many developers know, reaching a 1.0 milestone is huge, and something to
```
paulcsmith, line 18: Woo hoo!

paulcsmith (approved):
Lookin good! Stoked for this release. One high-level suggestion would be to move the "cool/new" stuff to the top: new ORM features, new Carbon features, etc. and the JS bundling and such lower down. Hook 'em with the new goodies. Especially helpful for people new to Lucky to know what's going on.

It may also be good to have a little blurb at the top about what Lucky has already done and what it is, in case people see the blog post and aren't totally sure what it is.

> Lucky is a fast, batteries included web framework built on the Crystal language. It catches bug with compile-time checks and aims to make things easy to maintain as the codebase grows. Lucky is already used in production, but with `rc1` we bring Lucky one step closer to a stable v1

Or something along those lines?

src/posts/lucky_1.0.0-rc1.cr
jwoertink: Ah, good catch. Avram is still included by default. I guess I was writing specifically to users that have existing Lucky apps in the wild :sweat_smile: Those users would have to manually add Avram back in.

src/posts/lucky_1.0.0-rc1.cr
```diff
@@ -0,0 +1,239 @@
...
+    `rc1`. We will be using this and the next few release candidates to clean up
+    the rough edges of the framework, and make sure all of the dependencies in the
+    Lucky ecosystem are all solid and ready to take us to the next level.
+
+    We will cover a few notable differences, but you can view all the changes
+    from the [CHANGELOG](https://github.com/luckyframework/lucky/blob/main/CHANGELOG.md)
+
+    ### How to upgrade
+
+    See the [UPGRADE_NOTES](https://github.com/luckyframework/lucky/blob/main/UPGRADE_NOTES.md#upgrading-from-029-to-030).
```
edwardloveall: I think you want 
```suggestion
    See the [UPGRADE_NOTES](https://github.com/luckyframework/lucky/blob/main/UPGRADE_NOTES.md#upgrading-from-030-to-100-rc1).
```

src/posts/lucky_1.0.0-rc1.cr
edwardloveall: This is so big. Great work on this!

edwardloveall (commented):
Nice work!

src/posts/lucky_1.0.0-rc1.cr
jwoertink: We removed turbolinks because the official repo is now archived by Rails. I didn't add in Turbo yet because there's a few questions I have unanswered like, do we need to do redirect stuff like Turbolinks had? Also, there's a few edge cases with Turbo where you need use a `<turbo-frame></turbo-frame>` tag (e.g. when an action redirects to itself), but what are the other cases? I just don't know enough about Turbo to add it confidently.

jwoertink: I guess I should also say that I'd be cool with adding Turbo in as a default, which is why I want people to try it out and see what needs to happen for that

src/posts/lucky_1.0.0-rc1.cr
jwoertink: Laravel-mix actually does this for us, and Vite also does this. Not really a "Lucky" specific thing, but just more part of that asset build system.

src/posts/lucky_1.0.0-rc1.cr
jwoertink: I'll open an issue on Carbon :+1:

src/posts/lucky_1.0.0-rc1.cr
jwoertink: actually it doesn't :joy: it should though lol

src/posts/lucky_1.0.0-rc1.cr
jwoertink: yeah, I just copied the code from Avram :joy:

src/posts/lucky_1.0.0-rc1.cr
jwoertink: Not really.. it's going to be adapter specific since the response is `settings.adapter.deliver_now(self)`. So like for sendgrid it'll be the client response from the API, but for SMTP would be a tuple with the status code and message...

jwoertink:
Thanks for the reviews @paulcsmith @edwardloveall @matthewmcgarvey :tada:

src/posts/lucky_1.0.0-rc1.cr
paulcsmith: That makes total sense to me! Can't really add it if we can't be confident we've got it all wired up just right. Turbo can be tricky to test all the edge cases

src/posts/lucky_1.0.0-rc1.cr
paulcsmith: Awesome!!

PR #1090, opened 2022-10-09
Update internationalization docs for rosetta 0.9.0

wout:
I've just released Rosetta 0.9.0 with compatibility for Crystal 1.6.0 and Lucky 1.0.0-rc1. There's a slight change in how it's used with Lucky. Translations without interpolations also need to be called with `.t`. Before, I applied some patches to allow values to be used in Lucky HTML without calling `.t`. For example:

```crystal
def render
  h1 r(".page_title")
end
```

instead of:

```crystal
def render
  h1 r(".page_title").t
end
```

This worked for previous versions of Lucky, but now it doesn't because of some changes. I could make this work again, but I've had doubts about this approach because it does not always work. Like in custom validations or API responses, for example, but in other places as well.

It also wasn't always obvious where it would work and where it wouldn't. So, in the end, I decided it would be best to be explicit about it and always require `.t` to be called. One additional benefit is that the lib is now much more resilient to changes in Lucky.

This PR reflects those changes.

### DIFF
--- a/src/actions/guides/frontend/internationalization.cr
+++ b/src/actions/guides/frontend/internationalization.cr
@@ -326,7 +326,7 @@ class Guides::Frontend::Internationalization < GuideAction
       needs current_user : User
       # ...
       def page_title
-        r(".page_title")
+        r(".page_title").t
       end
 
       def render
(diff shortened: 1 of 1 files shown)


jwoertink (approved):
Cool. Yeah, I think that's good. Thanks for the update!

PR #1108, opened 2022-10-30
Doc: mark postgres as optional (no more db dependency)

stan-slas:
### What
* Mark Postgres as optional in the guide [Getting started > Installing](https://luckyframework.org/guides/getting-started/installing)
* Add some docker information
* Typo fix in the blog

### Why
Sync the documentation with 1.0RC1 release. No more db dependency :partying_face: 
https://luckyframework.org/blog/lucky-1_0_0-rc1-release

When I started lucky I followed this documentation and I was afraid of installing Postgres on my host.
So I started to do my own containers at that moment ... and only after I realized:
* docker compose are already created for me
* db is not mandatory in lucky

I think it is worth mentioning it in the doc because it could be a deal breaker for some people.

### Screenshot
Before
image
After
image

### DIFF
--- a/src/actions/guides/getting-started/installing.cr
+++ b/src/actions/guides/getting-started/installing.cr
@@ -177,7 +177,13 @@ class Guides::GettingStarted::Installing < GuideAction
     This should return `#{LuckyCliVersion.current_version}`
 
     #{permalink(ANCHOR_POSTGRESQL)}
-    ## Postgres database
+    ## Postgres database (optional)
+
+    > You can skip this if you don't need database. Many sites (including this one) have no need for a database. In some cases,
+    your data comes from a 3rd party API, or maybe you need a custom database engine other than PostgreSQL.
+    
+    > You can skip this if you plan to use Lucky with a Postgres docker container. NB: A docker-compose configuration (with Postgres included) 
+    is provided out of the box when you start your project with `lucky init`.
 
     ### 1. Install Postgres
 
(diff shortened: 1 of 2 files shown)


src/actions/guides/getting-started/installing.cr
```diff
@@ -177,7 +177,13 @@ class Guides::GettingStarted::Installing < GuideAction
     This should return `#{LuckyCliVersion.current_version}`
 
     #{permalink(ANCHOR_POSTGRESQL)}
-    ## Postgres database
+    ## Postgres database (optional)
+
+    > You can skip this if you don't need database. Many sites (including this one) have no need for a database. In some cases,
+    your data comes from a 3rd party API, or maybe you need a custom database engine other than PostgreSQL.
+
+    > You can skip this if you plan to use Lucky with a Postgres docker container. NB: A docker-compose configuration (with Postgres included)
```
jwoertink: What's `NB:`?

jwoertink (commented):
I think this makes sense :+1:  Thanks

src/actions/guides/getting-started/installing.cr
stan-slas: You're right, I have the bad habit to put latin shortcuts everywhere :see_no_evil: 

Addressed here: https://github.com/luckyframework/website/pull/1108/commits/63c308fd4d74a139f4d82c043fd4be1c3c652ea3

I am not a native English speaker so feel free to suggest me any change :)
BTW: It means something like _Note well_ https://en.wikipedia.org/wiki/Nota_bene: pay attention, please notice, etc...

jwoertink: Ah, cool. Yeah, I hadn't seen that before. Thanks for the update!

jwoertink (approved):

PR #1111, opened 2022-11-05
Fix ubuntu deploy nginx

abbsi:
First PR :)

- The paths in instruction to create a config link from `sites-available` to `sites-enabled` were incorrect and missing the `ngnix`
- In the ngnix config, repalced `<yourapp>` with `<yourdomain>` under the SSL `server_name` directive. Without fix, config works fine if it's single site on server, but I believe it needs to be domain name if hosted along other sites to resolve correctly
- Clarified instructions on dealing with `default_server`


jwoertink (approved):
Looks good :+1:  Thanks for helping out!

PR #1152, opened 2023-02-02
Adding new guide for deploying to Fly.io

### LINKED ISSUE #1053: Add Fly.io deploy docs
Fly actually has really great docs already https://fly.io/docs/getting-started/crystal/

I think what we can do is under the deploying guide https://luckyframework.org/guides/deploying/heroku we can just have a link that says "Deploying to Fly.io", and have that link just click externally.

jwoertink:
Fixes #1053


matthewmcgarvey (approved):
Looks good but there's an issue with formatting

PR #1171, opened 2023-03-15
Updating Heroku to the latest build stack

jwoertink:


jwoertink:
ok.. Looks like I actually can't do this now. Heroku seems to have changed how these deploys work, and the pipeline config isn't available to me

image

Maybe @paulcsmith has access?? Maybe we should just look at moving to a different host? No clue... I'm gonna just merge this and YOLO :zany_face:

paulcsmith:
Unfortunately Heroku seems to have gone downhill quite a bit. I think Heroku is fine for now but I'd totally support moving elsewhere at this point. They have had a lot of security/downtime issues lately :(

PR #1190, opened 2023-04-05
Fix document

zw963:
Hi,  here is a local PR for try to fix some document issue when i am reading lucky document, some of them fixed by #1185, please have a look.

Following is some explain:

1.  i never saw a method named `render_flash`?

3.    `` The `path` method takes an `anchor` argument.``, where is the path method?

### DIFF
--- a/src/actions/guides/http_and_routing/flash.cr
+++ b/src/actions/guides/http_and_routing/flash.cr
@@ -27,9 +27,8 @@ class Guides::HttpAndRouting::Flash < GuideAction
     ```
 
     These will be rendered by the flash component in
-    `src/components/shared/flash_messages.cr`. The flash component is called from
-    your `MainLayout` and `AuthLayout` (found in `src/pages/`) by the
-    `render_flash` method.
+    `src/components/shared/flash_messages.cr`. The flash component was mounted from
+    your `MainLayout` and `AuthLayout` (found in `src/pages/`).
 
     You can modify the layout or the component to add HTML classes, change where
     flash is rendered, etc.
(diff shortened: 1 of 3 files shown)


src/actions/guides/http_and_routing/link_generation.cr
```diff
@@ -59,7 +59,7 @@ class Guides::HttpAndRouting::LinkGeneration < GuideAction
     In a web page, an anchor is a an element you can use to take a user to a specific part of a page.
     For example, we use this to link to different parts of a guide page.
 
-    The `path` method takes an `anchor` argument.
+    The `with` method takes an `anchor` argument.
```
jwoertink, line 62: Actually, both of these are true... :thinking:  I guess `with` is better here since we use it below, but you can pass `anchor` to the `path` method like here https://github.com/luckyframework/website/blob/2f11bb43ee55c332dca39b26074718c5c846ec7a/src/actions/guides/frontend/rendering_html.cr#L312

src/actions/guides/http_and_routing/flash.cr
```diff
@@ -27,9 +27,8 @@ class Guides::HttpAndRouting::Flash < GuideAction
     ```
 
     These will be rendered by the flash component in
-    `src/components/shared/flash_messages.cr`. The flash component is called from
-    your `MainLayout` and `AuthLayout` (found in `src/pages/`) by the
-    `render_flash` method.
+    `src/components/shared/flash_messages.cr`. The flash component was mounted from
+    your `MainLayout` and `AuthLayout` (found in `src/pages/`).
```
jwoertink, line 31: Good catch :+1:

jwoertink (approved):
Thanks for fixing these!

PR #1192, opened 2023-04-10
Fix incorrect AuthLayout example in i18n guide

bobmanary:
This fixes a compilation error caused by an incorrect key encountered while following the internationalization guide.


jwoertink (approved):
Good catch!

PR #1218, opened 2023-07-12
Add live reload code only in development

snadon:


src/components/shared/layout_head.cr
```diff
@@ -12,7 +12,7 @@ class Shared::LayoutHead < BaseComponent
       rss_link
 
       responsive_meta_tag
-      live_reload_connect_tag
+      live_reload_connect_tag if LuckyEnv.development?
```
jwoertink, line 15: This tag actually only renders anything with a specific compile-time flag. I guess maybe that logic is a bit hidden... :thinking: 

https://github.com/luckyframework/lucky/blob/22fcb187c2045a36823a03310d1ed68aec5d8f0b/src/lucky/tags/live_reload_tag.cr#L3-L12

snadon: Oh I see. I generated a new lucky project and found this line so I thought it was forgotten:

```
% rg live_reload
src/components/shared/layout_head.cr
15:      live_reload_connect_tag if LuckyEnv.development?
```

Then I rapidly checked the `app.js` compiled source and found:
```
?new WebSocket(this.src):new EventSource(this.src)
```

But I suppose it's coming from turbo.

jwoertink: Did you mean to make this PR on LuckyCLI? :joy:

snadon: Haha closing this in favor of a LuckyCLI PR.

PR #1219, opened 2023-07-12
Explicit images width and height

snadon:
Lighthouse told me to use explicit width and height to reduce layout shifts and improve CLS.

Also fixed a lint performance issue.


jwoertink (approved):
These changes seem ok. I wonder how they might affect responsive design... I'd imagine these are fine on mobile too, right?

snadon:
I used the responsive design mode of Safari as well as on Brave and it was fine. I didn't test it on an actual device though.

PR #1239, opened 2023-09-10
add mention of using :now in Time fields.

### LINKED ISSUE #1214: Document using `:now` as a default Time value
This is for migrations. You can do this:

```
add time : Time, default: :now
```

jwoertink:
Fixes #1214

This was a little tough because it's an edge case feature. and there's no sections dedicated to this sort of thing.

### DIFF
--- a/src/actions/guides/database/migrations.cr
+++ b/src/actions/guides/database/migrations.cr
@@ -164,6 +164,7 @@ class Guides::Database::Migrations < GuideAction
       add login_count : Int32, default: 0
       add tags : Array(String)
       add preferences : JSON::Any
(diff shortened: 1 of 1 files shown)


src/actions/guides/database/migrations.cr
```diff
@@ -164,6 +164,7 @@ class Guides::Database::Migrations < GuideAction
       add login_count : Int32, default: 0
       add tags : Array(String)
       add preferences : JSON::Any
+      add key : String, index: true
```
jwoertink, line 167: This isn't really related, but I wanted to sneak this in anyway to give an example that this is possible.

PR #1289, opened 2024-03-31
Rename Log to LogEntry in Logs::Index example

rmarronnier:
Because `Log` is used by Crystal std lib (https://crystal-lang.org/api/1.11.2/Log.html) you can't declare an Avram model like :

```crystal
class Log < BaseModel
  table do
    # You will define columns here. For example:
    # column name : String
  end
end
```

You'll get a compile time error.

One of the example of the website uses `LogQuery` and thus misleads you can have a model named simply `Log`.

I got bit by this in one of my project. One solution I found was to use the `LogEntry` name instead.

Open to any other name suggestions.


jwoertink:
oh, good call :raised_hands:

PR #1290, opened 2024-04-13
Update 02_assocations.cr

wweatherall:
The solution to an issue I had with my fortunes not giving user ids to the database, which requires user ids


jwoertink (approved):
Thanks for catching this :+1:

PR #1308, opened 2024-04-22
Adding new blog post to announce Lucky v1.2.0

### LINKED ISSUE #1301: Add blog post for v1.2.0 release
Using this changelog https://github.com/luckyframework/lucky/blob/main/CHANGELOG.md

jwoertink:
Fixes #1301

### DIFF
--- a/src/posts/lucky_1.2.0.cr
+++ b/src/posts/lucky_1.2.0.cr
@@ -10,7 +10,7 @@ class Lucky120Release < BasePost
     <<-TEXT
     We're well in to 2024 now. Crystal 1.12 is out,
     and so is Lucky 1.2.0. Let's check out all
-    of the great new features and updates
+    of the great new features and updates.
     TEXT
   end
 
@@ -46,7 +46,7 @@ class Lucky120Release < BasePost
     that way has been deprecated in favor of using `DB::Serializable`. This now keeps things consistent with `JSON::Serialzable` and
     `YAML::Serializable`, but it does come with a small breaking change.
 
-    The `DB::Serializable` is a module that is included in `Avram::Model`. It works by mapping the instance variables to columns
+    `DB::Serializable` is a module that is included in `Avram::Model`. It works by mapping the instance variables to columns
     in the database. This also means that any instance variable you've defined as custom will be mistaken for a database related column.
     To avoid this, you will need to add the annotation `@[DB::Field(ignore: true)]` to any instance variable you may have added
     in your model.
@@ -75,10 +75,10 @@ class Lucky120Release < BasePost
     # Find all users that have a username length of 8
     UserQuery.new.username.length.eq(8)
 
-    # Find all cards with a palindrome
+    # Find all cards with this palindrome
     CardQuery.new.title.reverse.eq("tacocat")
 
-    # Find all purchases where the amount rounded up is greater than 5
+    # Find all purchases where the rounded-up amount is greater than 5
     PurchaseQuery.new.amount.ceil.gt(5)
 
     # Turn a negative number in to positive, and use in your query
@@ -127,7 +127,7 @@ class Lucky120Release < BasePost
     end
     ```
 
-    You can now ignore the index created on `add_belongs_to` by specifying `index: false`.
+    You can also ignore the index created on `add_belongs_to` by specifying `index: false`.
 
     ```crystal
     alter table_for(Reaction), if_exists: true do
@@ -136,7 +136,7 @@ class Lucky120Release < BasePost
     end
     ```
 
-    This will allow you to create a custom unique index utilizing multiple columns instead of separate indexes on each column.
+    This will allow you to create a custom, unique index utilizing multiple columns instead of separate indexes on each column.
 
     We've also added the `create_sequence` and `drop_sequence` helpers.
 
@@ -201,8 +201,7 @@ class Lucky120Release < BasePost
 
     ### even more!
 
-    There were quite a few updates in this release, and unfortunately we can't show them all. Here's several
-    more that were added just for an encore!
+    Beyond the major changes listed above, there were quite a few updates in this release:
 
     * Support for Array(Enum) columns
     * Support for String primary keys
--- a/src/posts/lucky_1.2.0.cr
+++ b/src/posts/lucky_1.2.0.cr
@@ -26,7 +26,6 @@ class Lucky120Release < BasePost
     See the [UPGRADE_NOTES](https://github.com/luckyframework/lucky/blob/main/UPGRADE_NOTES.md#upgrading-from-110-to-120).
 
     You can also view the changes using [LuckyDiff](https://luckydiff.com/?from=1.1.0&to=1.2.0).
-    This tool helps see what has changed between versions and will be included in upgrade notes from now on.
 
     ## Here's what's new
 
@@ -42,8 +41,8 @@ class Lucky120Release < BasePost
     ### DB.mapping replaced with DB::Serializable
 
     Avram models use [crystal-db](https://github.com/crystal-lang/crystal-db/) under the hood to build the
-    columns. Previously we used `DB.mapping` to construct the columns, but as of [this commit](https://github.com/crystal-lang/crystal-db/),
-    that way has been deprecated in favor of using `DB::Serializable`. This now keeps things consistent with `JSON::Serialzable` and
+    columns. Previously we used `DB.mapping` to construct the columns, but as of [this PR](https://github.com/crystal-lang/crystal-db/pull/196),
+    that way has been deprecated in favor of using `DB::Serializable`. This is now more consistent with `JSON::Serialzable` and
     `YAML::Serializable`, but it does come with a small breaking change.
 
     `DB::Serializable` is a module that is included in `Avram::Model`. It works by mapping the instance variables to columns
@@ -89,19 +88,19 @@ class Lucky120Release < BasePost
     can be a bit tricky, and Avram now allows you to avoid a few unsafe raw WHERE queries by using type-safe ones.
 
     ```crystal
-    # where the preferences jsonb has the key 'theme'
+    # Where the preferences jsonb has the key 'theme'
     UserQuery.new.preferences.has_key("theme")
 
-    # where the preferences jsonb has the key 'theme' or the key 'color'
+    # Where the preferences jsonb has the key 'theme' or the key 'color'
     UserQuery.new.preferences.has_any_keys(["theme", "color"])
 
-    # where the preferences jsonb has the key 'theme' and the key 'color'
+    # Where the preferences jsonb has the key 'theme' and the key 'color'
     UserQuery.new.preferences.has_all_keys(["theme", "color"])
 
-    # where the preferences jsonb contains the JSON {"theme": "dark"}
+    # Where the preferences jsonb contains the JSON {"theme": "dark"}
     UserQuery.new.preferences.includes({theme: "dark"})
 
-    # where the preferences jsonb is found in this json document
+    # Where the preferences jsonb is found in this json document
     UserQuery.new.preferences.in(large_json_document)
     ```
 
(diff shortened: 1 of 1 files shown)


src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+
+  def published_at : Time
+    Time.utc(year: 2024, month: 4, day: 24)
+  end
+
+  def summary : String
+    <<-TEXT
+    We're well in to 2024 now. Crystal 1.12 is out,
+    and so is Lucky 1.2.0. Let's check out all
+    of the great new features and updates
```
edwardloveall:
```suggestion
    of the great new features and updates.
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+
+    Read through the [CHANGELOG](https://github.com/luckyframework/lucky/blob/main/CHANGELOG.md)
+    for all of the changes, but we will discuss some of the highlights here.
+
+    ### How to upgrade
+
+    See the [UPGRADE_NOTES](https://github.com/luckyframework/lucky/blob/main/UPGRADE_NOTES.md#upgrading-from-110-to-120).
+
+    You can also view the changes using [LuckyDiff](https://luckydiff.com/?from=1.1.0&to=1.2.0).
+    This tool helps see what has changed between versions and will be included in upgrade notes from now on.
```
edwardloveall: You can probably remove this sentence now since it's been in at least the past 6 blog posts.

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+    by installing Lucky on Windows via [Scoop](https://scoop.sh/).
+
+    This is a huge step forward for Crystal on Windows in general. As Crystal's support for Windows grows, we
+    will need to ensure the entire ecosystem is right there. You can check out our official [scoop bucket](https://github.com/luckyframework/scoop-bucket)
+    for instructions on installing, or to help contribute.
+
+    ### DB.mapping replaced with DB::Serializable
+
+    Avram models use [crystal-db](https://github.com/crystal-lang/crystal-db/) under the hood to build the
+    columns. Previously we used `DB.mapping` to construct the columns, but as of [this commit](https://github.com/crystal-lang/crystal-db/),
```
edwardloveall: This link does not go to a commit.

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+    that way has been deprecated in favor of using `DB::Serializable`. This now keeps things consistent with `JSON::Serialzable` and
```
edwardloveall:
```suggestion
    that API has been deprecated in favor of using `DB::Serializable`. This new API is consistent with `JSON::Serialzable` and
```

I'm not actually sure it's consistent. Check me on that.

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+    `YAML::Serializable`, but it does come with a small breaking change.
+
+    The `DB::Serializable` is a module that is included in `Avram::Model`. It works by mapping the instance variables to columns
```
edwardloveall:
```suggestion
    `DB::Serializable` is a module that is included in `Avram::Model`. It works by mapping the instance variables to columns
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+    the column with `UPPER(col)` or `LOWER(col)` which can be used for special comparrisons.
+
+    In this release, we have added a LOT more. The `String` type added `length` and `reverse`.
+    `Int` and `Float` types received a few `abs`, `ceil`, and `floor`.
+
+    ```crystal
+    # Find all users that have a username length of 8
+    UserQuery.new.username.length.eq(8)
+
+    # Find all cards with a palindrome
```
edwardloveall:
```suggestion
    # Find all cards with this palindrome
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+    CardQuery.new.title.reverse.eq("tacocat")
+
+    # Find all purchases where the amount rounded up is greater than 5
```
edwardloveall:
```suggestion
    # Find all purchases where the rounded-up amount is greater than 5
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+
+    # Turn a negative number in to positive, and use in your query
+    PointQuery.new.x_position.abs.lte(1)
+    ```
+
+    To add to these great methods, we've also added `JSON::Any` critera methods. Querying jsonb fields
+    can be a bit tricky, and Avram now allows you to avoid a few unsafe raw WHERE queries by using type-safe ones.
+
+    ```crystal
+    # where the preferences jsonb has the key 'theme'
```
edwardloveall: You capitalized the last set of comments. Maybe these too?

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+    with `if_not_exists` and `if_exists` respectively.
+
+    ```crystal
+    # Only create the table if it does not already exist
+    create table_for(Server), if_not_exists: true do
+      add hostname : String
+    end
+    ```
+
+    You can now ignore the index created on `add_belongs_to` by specifying `index: false`.
```
edwardloveall:
```suggestion
    You can also ignore the index created on `add_belongs_to` by specifying `index: false`.
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+
+    ```crystal
+    alter table_for(Reaction), if_exists: true do
+      add_belongs_to user : User?, index: false
+      add_belongs_to friend : User?, index: false
+    end
+    ```
+
+    This will allow you to create a custom unique index utilizing multiple columns instead of separate indexes on each column.
```
edwardloveall:
```suggestion
    This will allow you to create a custom, unique index utilizing multiple columns instead of separate indexes on each column.
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+    def rollback
+      drop_sequence table_for(Account), :number
+    end
+    ```
+
+    ### The IGNORE constant
+
+    In Avram, the attributes defined on operations are a union of the type you specified, and `Avram::Nothing`.
+    The `Avram::Nothing` type is used as a way to differentiate between you intending to update a column to `nil`
+    by passing in a `nil` value versus a `nil` value existing in params but with no intention of updating the column.
```
edwardloveall: This was really confusing and I did my best. Is this correct? Is there something that I'm leaving out that's important here? I can't find any existing docs to check my knowledge here.

```suggestion
      The `Avram::Nothing` type was used when updating a model to specify that a column should be set to `NULL`. This is instead of ignoring that column in the update statement entirely.
```

It may also help to show a before and after.

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+      def cake
+        {
+          io: @image,
+          file_name: "birthday_cake.png",
+          mime_type: "image/png"
+        }
+      end
+
+      def after_send(result)
+        @image.close
```
edwardloveall: Do devs need to remember to close the file? If so, that would be a good thing to mention, even in a comment:

```suggestion
        # remember to close your open files!
        @image.close
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,237 @@
...
+      end
+    end
+    ```
+
+    ### even more!
+
+    There were quite a few updates in this release, and unfortunately we can't show them all. Here's several
+    more that were added just for an encore!
```
edwardloveall:
```suggestion
    Beyond the major changes listed above, there were quite a few updates in this release:
```

edwardloveall (commented):
Nice work! Tons of stuff here.

src/posts/lucky_1.2.0.cr
jwoertink: oops! :joy:

src/posts/lucky_1.2.0.cr
jwoertink: This is what I mean

```crystal
class Foo
  include YAML::Serializable
  property bar : String
end

class Foo
  include JSON::Serializable
  property bar : String
end

class Foo
  include DB::Serialiable
  property bar : String
end

# VS

class Foo
  DB.mapping({
    bar: {
      type: String
    }
  })
end
```

src/posts/lucky_1.2.0.cr
jwoertink: That's not quite correct. The `Avram::Nothing` type didn't go away. `IGNORE` is just defined as `IGNORE = Avram::Nothing.new`. It's used when you are passing in `nil` but you *don't* want to set the value to `NULL`. The real benefit here is just that instead of having 1000 instances of `Avram::Nothing`, you now have 1. Does that clarify better?

src/posts/lucky_1.2.0.cr
jwoertink: I didn't implement the attachments, so I wasn't sure how this worked, to be honest. They take an `IO` object, but `File.read()` returns a `String`, so I just did this as a shot in the dark :joy: I guess I could swap to using `IO.copy` then the file.close is irrelevant. Would that be better for this example?

src/posts/lucky_1.2.0.cr
jwoertink: Reworded this, so maybe it's more clear now.

src/posts/lucky_1.2.0.cr
jwoertink: Reworded this. It's a bit complicated to explain, but let me know if that reads better.

src/posts/lucky_1.2.0.cr
jwoertink: I'm now using IO.copy in the example. It's probably better anyway because I think if the email fails to send, the `after_send` isn't called :thinking:

src/posts/lucky_1.2.0.cr
edwardloveall: Love it.

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,242 @@
...
+    ```
+
+    ### The IGNORE constant
+
+    In Avram, the attributes defined on operations are a union of the type you specified, and `Avram::Nothing`.
+    The `Avram::Nothing` type is used as a way to differentiate `nil` values passed through params that are
+    intended to update columns to a `NULL` value versus values you wish to ignore.
+
+    Previously, each attribute would define a default value of `Avram::Nothing.new`. The larger your application grew,
+    the more instances of `Avram::Nothing` you would have. These all have now been replaced with a single instance of `Avram::Nothing`
```
edwardloveall:
```suggestion
    the more instances of `Avram::Nothing` you would have. These have all been replaced with a single instance of `Avram::Nothing`
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,242 @@
...
+    UpdateUserEmail.update!(
+      current_user,
+      email: new_email.presence || IGNORE
+    )
+    ```
+
+    ### Email attachments
+
+    [Carbon](https://github.com/luckyframework/carbon/pull/88) received a nice update with the addition of
+    attachment support.
```
edwardloveall: Technically you're not linking to Carbon here, but the update to carbon

```suggestion
    Carbon received a nice update that adds of [attachment support](https://github.com/luckyframework/carbon/pull/88).
```

src/posts/lucky_1.2.0.cr
```diff
@@ -0,0 +1,242 @@
...
+    def rollback
+      drop_sequence table_for(Account), :number
+    end
+    ```
+
+    ### The IGNORE constant
+
+    In Avram, the attributes defined on operations are a union of the type you specified, and `Avram::Nothing`.
+    The `Avram::Nothing` type is used as a way to differentiate `nil` values passed through params that are
+    intended to update columns to a `NULL` value versus values you wish to ignore.
```
edwardloveall:
```suggestion
    Passing `nil` during an operation sets a value to `NULL` in the database. In contrast, using `Avram::Nothing` ignores the value; the value is unchanged.
```

PR #1366, opened 2024-08-15
Improvements to Postgresql documentation

sharkos:
I have written some additonal steps for setting up PostgreSQL on Linux. improvements include:

- workaround for Lucky 1.20 bug with database create on Linux which requires addition of a local database named the same as the user specified in the databse config
- added notes on creating a local database user with the 'createuser' command included with postgres
- modified reference of Postgresql documentation to v14
- replaced references to using the builtin 'postgres' user with a user called 'lucky', it is best practice to never use the built in sa or superuser account of a database
- minor notes and changes to improve readability.

### DIFF
--- a/src/actions/guides/database/database_setup.cr
+++ b/src/actions/guides/database/database_setup.cr
@@ -28,8 +28,8 @@ class Guides::Database::DatabaseSetup < GuideAction
 
     * database : `String` - This is the name of your database. The default is set at the top of this file.
     * hostname : `String?` - The host where your database is located. Generally "localhost".
-    * username : `String?` - Your database user.
-    * password : `String?` - Your password.
+    * username : `String?` - Your database user. Our example uses a database user "lucky".
+    * password : `String?` - Your password. (Caution: do not store production passwords in your src repo)
     * port : `Int32?` - The port to connect to. Default is `5432`
     * query : `String?` - A query string of connection pool settings.
 
@@ -40,7 +40,8 @@ class Guides::Database::DatabaseSetup < GuideAction
         database: database_name,
         hostname: "localhost",
         port: 5432,
-        username: "postgres",
+        username: "lucky",
+        password: "lucky",
         query: "initial_pool_size=5&retry_attempts=2"
       )
     end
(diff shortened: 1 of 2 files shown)


src/actions/guides/database/database_setup.cr
```diff
@@ -28,8 +28,8 @@ class Guides::Database::DatabaseSetup < GuideAction
 
     * database : `String` - This is the name of your database. The default is set at the top of this file.
     * hostname : `String?` - The host where your database is located. Generally "localhost".
-    * username : `String?` - Your database user.
-    * password : `String?` - Your password.
+    * username : `String?` - Your database user. Our example uses a database user "lucky".
+    * password : `String?` - Your password. (Caution: do not store production passwords in your src repo)
```
jwoertink, line 32: :+1:

src/actions/guides/getting-started/installing.cr
```diff
@@ -268,6 +272,51 @@ class Guides::GettingStarted::Installing < GuideAction
...
+    sudo su - postgres
+    ```
+
+    Now create the user, and set the roles `createrole`, `superuser`, `createdb` then enter a password. In
+    this example below, we will create a user named 'lucky' and set a password of 'lucky'.
+    ```bash
+    createuser -dsrP lucky
+    ```
+
+    > Due to a bug in Lucky version 1.20, you will have to create a database with the same name as your
```
jwoertink: Technically not a bug, and it was introduced in 1.1.0. This was intentional, it was just a poor choice :rofl: 

I think the wording is fine, but the version should say `1.2.0` at least

jwoertink (changes requested):
Thanks so much for adding this in. It's a lot more clear for new users and will be a huge help. There's just one small change in the version, but I think it's all good to go :+1:

jwoertink:
Looks like there's some formatting that needs to be fixed too.

src/actions/guides/getting-started/installing.cr
sharkos: OK, I will make some wording adjustments and fix the version string.

jwoertink (approved):
:raised_hands:

Issue #1081, opened 2022-09-29
Update `validate_numeric` docs


jwoertink:
the arg names are changed

https://github.com/luckyframework/avram/pull/867

jwoertink:
this may need to hold off https://github.com/luckyframework/avram/pull/867#issuecomment-1263683054

PR #1392, opened 2024-10-05
Update validate_numeric args

snadon:
Fixes for https://github.com/luckyframework/website/issues/1081


jwoertink (approved):
Thanks. I kept putting this off because there was a plan to change the arg names again, but no clue when I'll ever get to that so we might as well have this correct for now and worry about a change later. :sweat_smile:

PR #1478, opened 2025-06-16
Fix #1476

### LINKED ISSUE #1476: Document path_without_query_params
This is an action class method that lets you return the action's path without query params.

```crystal
class SomeAction < BrowserAction
  param page : Int32

  get "/path" do
    plain_text "ok"
  end
end

# This will raise because `page` is required
SomeAction.path

SomeAction.path_without_query_params # => "/path"
```

zw963:
And, both add https://crystal-china.org as awesome link


jwoertink (approved):
Thank you!

PR #1481, opened 2025-06-22
Blogs v1.4.x

jwoertink:
This is the blog post announcement for Lucky v1.4.x

### DIFF
--- a/src/models/lucky_cli_version.cr
+++ b/src/models/lucky_cli_version.cr
@@ -8,7 +8,7 @@ module LuckyCliVersion
   end
 
   def current_version : SemanticVersion
-    SemanticVersion.new(1, 4, 0)
+    SemanticVersion.new(1, 4, 1)
   end
 
   def min_compatible_crystal_version : SemanticVersion
--- a/src/posts/lucky_1.4.1.cr
+++ b/src/posts/lucky_1.4.1.cr
@@ -142,7 +142,7 @@ class Lucky141Release < BasePost
     ### State of LuckyFlow
 
     If you're not familiar, LuckyFlow is a shard that ships with new browser based Lucky apps that
-    helps with doing Acceptance testing by spinning up a browser and wallking through a user's flow
+    helps with doing Acceptance testing by spinning up a browser and walking through a user's flow
     using your site.
 
     As you can see, we went from v1.4.0 right to v1.4.1. This was due to some bugs we found in LuckyFlow
@@ -154,9 +154,9 @@ class Lucky141Release < BasePost
     quite complex. On top of this, needing to support these drivers across different platforms is also
     very complex.
 
-    For the time being, generating a new Lucky app will still sihp with LuckyFlow, but the specs will be
+    For the time being, generating a new Lucky app will still ship with LuckyFlow, but the specs will be
     set to "pending" until we can release a new version with the issues fixed. You are still free to use
-    LuckyFlow, but just keep in mind that your specs may radomly fail, and the failure may be unrelated
+    LuckyFlow, but just keep in mind that your specs may randomly fail, and the failure may be unrelated
     to your code.
 
     > Follow along with the discussion on [Github](https://github.com/luckyframework/lucky_cli/issues/890)
--- a/src/posts/lucky_1.4.1.cr
+++ b/src/posts/lucky_1.4.1.cr
@@ -48,7 +48,7 @@ class Lucky141Release < BasePost
     end
     ```
 
-    As you can see, this will limit the number of requests to the endpoint to a max of 1 every 1 minute. It will use the
+    This will limit the number of requests to the endpoint to a max of 1 every 1 minute. It will use the
     `rate_limit_identifier` key for the lookup which you can customize if you need to set the limit to different amounts
     on a per user basis, for example.
 
@@ -90,12 +90,12 @@ class Lucky141Release < BasePost
 
     ### Table Locking
 
-    In general, a table lock is a way to lock your table while you perform operations on it preventing other connections
-    from altering that table during that time. In Postgres, however, there's 8 different types of locking that can
-    allow for a lot more customization and handling when it comes to locking your table.
+    Table locks control which actions can be performed on a table at the same time. Postgres has
+    [eight different table locking modes](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-TABLES) that
+    allow you to specify the locking behavior. Avram now supports all eight.
 
     The `AppDatabase.with_lock_on` method takes the model whose table you will be locking, and the locking `mode`.
-    Then within the block, all queries made will happen within the transaction.
+    Then within the block, the lock is applied and all queries made will happen within a transaction.
 
     ```crystal
     AppDatabase.with_lock_on(User, mode: :row_exclusive) do
@@ -104,7 +104,7 @@ class Lucky141Release < BasePost
     end
     ```
 
-    Avram supports the following lock modes:
+    These are the eight modes:
 
     ```
     ACCESS_SHARE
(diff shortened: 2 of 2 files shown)


src/models/lucky_cli_version.cr
```diff
@@ -8,7 +8,7 @@ module LuckyCliVersion
   end
 
   def current_version : SemanticVersion
-    SemanticVersion.new(1, 4, 0)
+    SemanticVersion.new(1, 4, 1)
```
edwardloveall, line 11: Just an FYI that the PR title is 1.4.0 but the version is 1.4.1

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+    ### Rate Limiting
+
+    Lucky Actions now have a "Rate Limiting" feature that allows you to limit the number of calls to an action within
+    a specified period. To use this feature, you will just need to include the new `Lucky::RateLimit` module.
+
+    ```crystal
+    class Users::LivePingCheck < ApiAction
+      include Lucky::RateLimit
+
+      rate_limit to: 1, within: 1.minute
```
edwardloveall, line 39: Rad

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+        plain_text "ok"
+      end
+
+      private def rate_limit_identifier : String
+        "live-ping-check"
+      end
+    end
+    ```
+
+    As you can see, this will limit the number of requests to the endpoint to a max of 1 every 1 minute. It will use the
```
edwardloveall:
```suggestion
    This will limit the number of requests to the endpoint to a max of 1 every 1 minute. It will use the
```

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+    `rate_limit_identifier` key for the lookup which you can customize if you need to set the limit to different amounts
+    on a per user basis, for example.
+
+    ### No more bash script setups
+
+    In order for us to better support Windows, we needed to remove the bash scripts that were used for bootstrapping a
+    Lucky application. While using Powershell would have been an option, this also would have made the code in the LuckyCLI
+    much more difficult to maintain due to having different codebase structures for Windows and non-Windows.
+
+    Instead, we decided to just port all of the scripts straight to Crystal. This is now a lot more cross-platform compatible
```
edwardloveall, line 61: Smart.

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+      def self.write_database : Avram::Database.class
+        WriterDB
+      end
+    end
+
+    # Uses the ReaderDB
+    user = UserQuery.find(1)
+
+    # Uses the WriterDB
+    SaveUser.update!(user, last_active_at: Time.utc)
```
edwardloveall, line 85: Woah!

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+    ```
+
+    Currently this will only change the connection between doing queries and saving records through Save or Delete operations.
+    Handling tasks such as fallbacks, or retries based on latency from reading and writing are not supported.
+
+    ### Table Locking
+
+    In general, a table lock is a way to lock your table while you perform operations on it preventing other connections
+    from altering that table during that time. In Postgres, however, there's 8 different types of locking that can
+    allow for a lot more customization and handling when it comes to locking your table.
```
edwardloveall: How's this? Still getting what you were trying to say? I removed the "generic table locks bit" cut to the Postgres chase.

```suggestion
    Table locks control which actions can be performed on a table at the same time. Postgres has
    [eight different table locking modes](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-TABLES) that
    allow you to specify the locking behavior. Avram now supports all eight.
```

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+
+    The `AppDatabase.with_lock_on` method takes the model whose table you will be locking, and the locking `mode`.
+    Then within the block, all queries made will happen within the transaction.
```
edwardloveall: Is this true?

```suggestion
    Then within the block, the lock is applied and all queries made will happen within a transaction.
```

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+
+    ```crystal
+    AppDatabase.with_lock_on(User, mode: :row_exclusive) do
+      user = UserQuery.new.id(1).for_update.first
+      SaveUser.update!(user, name: "New Name")
+    end
+    ```
+
+    Avram supports the following lock modes:
```
edwardloveall:
```suggestion
    These are the eight modes:
```

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+    ```crystal
+    LuckyEnv.cache_enabled? #=> true
+    LuckyEnv.port #=> 3000
+    LuckyEnv.host #=> "localhost"
+    LuckyEnv.exchange_rate #=> 3.14
+    ```
+
+    ### State of LuckyFlow
+
+    If you're not familiar, LuckyFlow is a shard that ships with new browser based Lucky apps that
```
edwardloveall, line 144: Up to you, but I think it's fair to assume that people reading the release blog post know what LuckyFlow is, but maybe I'm projecting too much of my own knowledge.

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+    helps with doing Acceptance testing by spinning up a browser and walking through a user's flow
+    using your site.
+
+    As you can see, we went from v1.4.0 right to v1.4.1. This was due to some bugs we found in LuckyFlow
```
edwardloveall: You're also tee-ing up every paragraph here (e.g. "As you can see") and then saying what you came to say. Not a big deal, but you might try starting with the statement and adding the additional language only if it's necessary.

One way could be:

```suggestion
    We released LuckyFlow v1.4.1 and skipped v1.4.0. This was due to some bugs we found in LuckyFlow
```

src/posts/lucky_1.4.1.cr
```diff
@@ -0,0 +1,182 @@
...
+    that were causing specs to fail randomly on new generated apps. These failures are new and seem to be
+    race conditions that we have yet to find.
+
+    Due to the complexity of LuckyFlow needing several dependencies, like Selenium, and handling several
+    different webdrivers (e.g chromedriver, geckodriver, etc...), keeping LuckyFlow updated has become
+    quite complex. On top of this, needing to support these drivers across different platforms is also
+    very complex.
```
edwardloveall: Or here, starting with "keeping LuckyFlow updated has become quite complex" and the moving on to the "why".

edwardloveall (commented):
Kapow! Nice work Jeremy and everyone that contributed.

src/posts/lucky_1.4.1.cr
jwoertink: ah, yeah, I like that better. Thanks!

src/posts/lucky_1.4.1.cr
jwoertink: Fair. I was live streaming it while writing and just talking through it to people :joy:  It's probably fine though.

src/posts/lucky_1.4.1.cr
jwoertink: Good catch. I do that a lot :sweat_smile:

jwoertink:
Thanks for the review @edwardloveall

PR #1537, opened 2026-04-26
Updating the tutorial for Lucky 1.5.0

### LINKED ISSUE #1535: Update the tutorial
The tutorial needs a thorough look over with Lucky 1.5.0. I'm sure it still mentions yarn and all that.

jwoertink:
Fixes #1535

There's quite a few outdated things in the guides. I'm trying to stick mostly to the tutorial, but a few links in the tutorial go to other pages with bad information, so I've lumped a few of those in too.

Also, originally I went with Bootstrap in order to add in some styled components and baked in javascript stuff.... I'm playing around with changing that to something a little more lightweight just for the sake of simplicity.

### DIFF
--- a/src/actions/guides/command_line_tasks/built_in.cr
+++ b/src/actions/guides/command_line_tasks/built_in.cr
@@ -12,7 +12,7 @@ class Guides::CommandLineTasks::BuiltIn < GuideAction
     Lucky comes bundled with [LuckyTask](https://github.com/luckyframework/lucky_task) for creating tasks.
 
     Tasks are used to run some code that sits outside of your main application. When you need to add a task, you'll place it in the `tasks` folder.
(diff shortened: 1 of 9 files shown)


wout:
Slightly unrelated, there's one thing about the tutorial that I frequently run into. If you use the search (Avram stuff for example), often you land on the tutorial, which is a bit confusing.
I wonder if the tutorial could be filtered out, or could get a lower priority in the search or something.

jwoertink:
> I wonder if the tutorial could be filtered out, or could get a lower priority in the search

That's an interesting thought. I'll need to see how we can configure that, but it makes sense. You shouldn't really need to search stuff in the tutorial.

jwoertink:
For the algolia search, I just logged into the account and set an exclusion rule for `/tutorial`. We'll see how that works out.

